Issue #100, opened 2026-04-26
feat(cli): support multi-format output in a single run (--format json,markdown)


cmbays:
## Summary

Allow `crap4rs` to emit multiple output formats from a single invocation, e.g. `--format json,markdown` writing both an envelope and a rendered scorecard in one analysis pass.

## Motivation

The [scorecard composite action](https://github.com/breezy-bays-labs/crap4rs/blob/main/.github/actions/scorecard/action.yml) (PR #99) currently runs \`crap4rs\` twice — once for the JSON envelope, once for the markdown scorecard:

\`\`\`bash
crap4rs ... --format json     --no-fail > envelope.json
crap4rs ... --format markdown --no-fail > scorecard.md
\`\`\`

Each pass repeats the LCOV parse + syn walk + CRAP recompute. Cheap relative to runner setup, but a real waste in any aggregator pipeline that wants both shapes.

## Acceptance Criteria

- [ ] \`--format json,markdown\` (or similar) accepted by clap; produces two output streams
- [ ] Single analysis pass; reporters fan out from the same in-memory \`AnalysisResult\` + \`AnalysisDelta\`
- [ ] Output destination strategy decided: stdout multiplex vs. \`--output-file\` per format. Probably the latter — \`--json-output FILE --markdown-output FILE\` or \`--format json:envelope.json,markdown:scorecard.md\`
- [ ] Backwards compatible: existing single-format \`--format X\` flow unchanged

## Out of scope

- HTML/SARIF (separate issues #71/#70)

## Surfaced by

- PR #99 review (Gemini comment on action.yml two-pass invocation)

## Labels

\`type:feature\` \`priority:later\`

Issue #82, opened 2026-04-25
feat(domain): FunctionVerdict.diagnostic slot for /cut-the-crap agent


cmbays:
## Summary
Add `Option<Diagnostic>` slot on `FunctionVerdict` for the future `/cut-the-crap` agent skill — structured per-function refactor signals.

## Context
Once the agent skill ships, agents need machine-readable refactor guidance: contributor line ranges, suggested extractions, coverage-gap line lists, naming concerns. ADR D7 reserves this slot. Additive on the existing struct under `#[non_exhaustive]` — non-breaking when added later.

## Acceptance Criteria
- [ ] `FunctionVerdict.diagnostic: Option<Diagnostic>` field added (initial population in this issue or follow-up)
- [ ] `Diagnostic` struct is language-agnostic (lives in `domain::view` or `domain::diagnostic`)
- [ ] JSON envelope `view.shown[].diagnostic` is opt-in or always present (decide at impl)
- [ ] Cross-reference with `/cut-the-crap` agent skill spec when written

Pipeline: crap4rs/20260425-cli-presentation-view (V1a-V6 follow-up)
ADR: ~/Github/ops/decisions/crap4rs/adr-view-abstraction.md

Issue #77, opened 2026-04-25
feat: /cut-the-crap reference Claude Code skill


cmbays:
## Summary

Ship a reference Claude Code skill — `/cut-the-crap` — at `.claude/skills/cut-the-crap/SKILL.md` in the crap4rs repo. The skill consumes `--format advice` output (#76) and drives the cover-then-split remediation loop, naming proposed function splits with rationale.

## Motivation

crap4rs the *tool* should stay a sharp unix-style emitter (mechanical analysis, structured output). The *agent loop* — cover untested code first, then propose named extractions, write a plan, optionally open a PR — belongs in a skill, not the binary. Skills are user-shaped, swappable, and don't bloat the tool.

Shipping a reference skill in-repo lets any crap4rs user install it with one command:

    cp -r .claude/skills/cut-the-crap ~/.claude/skills/

This is the user-facing realization of Uncle Bob's CRAP-rant loop (2026-04-24 video): an agent you point at a codebase that methodically drives every function below threshold, explaining each cut.

## Acceptance Criteria

- [ ] `.claude/skills/cut-the-crap/SKILL.md` with frontmatter (`name`, `description`, `when_to_use`)
- [ ] Skill invokes `crap4rs --format advice` (depends on #76)
- [ ] For each over-threshold function, applies the cover-first heuristic:
  - `root_cause: low_coverage` → write tests covering `uncovered_lines`, re-run, confirm score drop
  - `root_cause: high_complexity` → propose names + rationale for each `proposed_splits` entry, then apply
  - `root_cause: both` → cover first, re-evaluate, then split if still over
- [ ] Skill emits a structured plan to `tmp/cut-the-crap-plan.md` before applying changes
- [ ] `--explain-only` mode: produce the named-splits plan for human review without applying
- [ ] Idempotent: re-running on already-passing functions is a no-op
- [ ] README documents installation and usage
- [ ] Reference example: dogfood the skill on crap4rs itself, link the resulting PR

## Blocked By

- #76 — `--format advice` (the skill consumes that JSON shape)

## Context

The naming + rationale layer is the skill's value-add — `--format advice` (#76) gives mechanical split candidates ("extract lines 47-62, branch path X, complexity contribution 8"); the skill names the result ("`validate_payment_window` — isolates the date-range guard so it can be tested independently of the dispatch logic").

Discovered during 2026-04-25 roadmap conversation triggered by Uncle Bob's "Morning Bathrobe Rant: CRAP" video.

Issue #76, opened 2026-04-25
feat: --format advice (structured remediation hints)


cmbays:
## Summary

Add `--format advice` (alias `--explain`) — for each over-threshold function, emit structured remediation hints: cover-first vs. extract-function, with mechanical AST-derived split candidates.

## Motivation

crap4rs already identifies *which* functions exceed the threshold, and #59 added breakdown contributors so users see *what* drives the score. The missing piece is *what to do about it* — in a structured shape an agent (or human) can act on without re-reading the source.

The remediation order is well-known (Uncle Bob CRAP framing, 2026-04-24 ""Morning Bathrobe Rant: CRAP"" video):

1. If coverage is low → add tests first (cheaper, often resolves the score)
2. If complexity is the actual driver → extract sub-functions

crap4rs has the AST, the complexity calc, and the per-function coverage. It can mechanically point at the *best* split candidates (deepest nesting, branch with largest complexity contribution, largest sub-block by line range). That's mechanical; an agent can take it from there.

**Crucially out of scope:** the tool does NOT name the splits or write rationale prose. That's an agent's job. The tool emits the structured input; the agent does the naming. Keeps the static-analysis layer honest and makes the consuming skill swappable.

## Acceptance Criteria

- [ ] `--format advice` emits JSON, one entry per over-threshold function with shape:

      {
        "function": "...",
        "file": "...",
        "line_range": [start, end],
        "crap_score": ...,
        "root_cause": "low_coverage" | "high_complexity" | "both",
        "suggested_action": "add_tests" | "extract_subfunctions" | "both",
        "uncovered_lines": [...],   // when add_tests is suggested
        "proposed_splits": [        // when extract_subfunctions is suggested
          {
            "line_range": [start, end],
            "complexity_contribution": N,
            "branch_path": "...",
            "kind": "deepest_nesting" | "largest_subblock" | "highest_branch_count"
          }
        ]
      }

- [ ] `--explain` is an alias of `--format advice`
- [ ] No naming, no prose rationale, no LLM-shaped guesses — purely AST-derived
- [ ] Schema documented in README under "Output formats"
- [ ] Composable with all filtering flags (#62, #63, #65, #68 quartet)

## Context

Designed as the structured input for the `/cut-the-crap` skill (separate issue, blocked by this) which consumes this format and drives the cover-then-split loop with named outputs.

[...]

PR #106, opened 2026-04-27
feat(domain): start_column/end_column on SourceSpan for SARIF region precision (#105)

### LINKED ISSUE #105: feat(domain): add start_column/end_column to SourceSpan for precise SARIF regions
## Summary

Extend `SourceSpan` with `start_column` / `end_column` so SARIF results can carry full `region` precision. Surfaced as a follow-up to #70 / PR #104 (Gemini bot review).

## Motivation

SARIF v2.1.0 `region` supports `startColumn` and `endColumn` alongside `startLine` / `endLine`. GitHub Code Scanning uses these to highlight the exact byte range in the PR diff rather than the entire line. The current SARIF reporter (#70) emits `startLine` and `endLine` only because `domain::types::SourceSpan` does not carry column data.

`syn::spanned::Spanned` exposes `Span::start().column` and `Span::end().column`, so the data is available — the complexity adapter just doesn't capture it at the function level today (it does at the contributor level via `ComplexityContributor.column`).

## Acceptance Criteria

- [ ] `SourceSpan` gains `start_column: usize` and `end_column: usize` (1-based per LSP / SARIF convention)
- [ ] `adapters::complexity` populates the new fields from `proc_macro2::Span`
- [ ] Existing line-range matching is unaffected (columns are additive metadata, not used for matching)
- [ ] `format_sarif` emits `region.startColumn` / `region.endColumn` when both are non-zero
- [ ] JSON envelope `schema_version` stays at 1 (additive change with `#[serde(default)]`)
- [ ] Existing snapshots regenerate cleanly (single expected diff per snapshot)

## Context

- Discovered via gemini-code-assist[bot] inline comment on PR #104
- SARIF v2.1.0 region spec: https://docs.oasis-open.org/sarif/sarif/v2.1.0/sarif-v2.1.0.html#_Toc34317685
- Relates to the unified `crap-core` extraction (CLAUDE.md): SourceSpan must stay language-agnostic — column semantics are universal (1-based UTF-16 code units per LSP), no Rust-specific assumptions

cmbays:
Closes #105

## Summary

Additive `start_column` / `end_column` fields on `domain::types::SourceSpan` (1-based inclusive; `0` signals "unknown") so SARIF regions can render column-precise underlines under GitHub Code Scanning. Follows up the L198 review on PR #104 (#70).

The Rust complexity adapter populates from `proc_macro2::Span::start()/end().column` (0-based; we add 1 to convert to 1-based for `start_column`, and the inclusive convention for `end_column` happens to match `sp.end().column` directly because proc_macro2 returns exclusive end). Diff-hunk adapter sets columns to `0` (line-only data — no fabricated values).

The SARIF reporter emits `region.startColumn` / `endColumn` only when **both** columns are nonzero (a half-known span is strictly worse than no column data). SARIF v2.1.0 §3.30.7 specifies `endColumn` as exclusive, so the reporter adds 1 at the wire boundary; SourceSpan stays consistent with the inclusive `end_line` convention.

## Test plan

- [x] `cargo fmt --check && cargo clippy --all-targets -- -D warnings` clean
- [x] `cargo nextest run` — 757/757 pass (was 751 before — added 4 unit + 1 integration test, plus 1 baseline existing column test)
- [x] Self-CRAP gate: new functions stay well under strict 15:
  - `span_of` CC=1 (linear), `region_for_span` CC=2 (single conditional)
- [x] Mutation gate scoped to new code paths: `cargo mutants` initially surfaced 3 survivors on the `+ 1` arithmetic — all three are now caught by deterministic in-source tests pinning EXACT column values (`fn foo() {}` → `start=1, end=11`; `    pub fn bar() {}` → `start=5, end=19`).
- [x] Manual smoke: `crap4rs --coverage lcov.info --threshold 15 --format sarif` against crap4rs's own source emits a finding for `cli/mod.rs::run_inner` with:
  ```json
  {
    "startLine": 564,
    "endLine": 732,
    "startColumn": 1,
    "endColumn": 2
  }
  ```
  both columns 1-based, sensible for a top-level `pub fn`.

## Snapshot impact

- `full_json_snapshot.snap` — additive `start_column: 0, end_column: 0` per span (4 spans affected). All test fixtures use synthetic spans (no column data), so they emit `0`.
- `full_sarif_snapshot.snap` — unchanged. The multi-function fixture spans all have column 0, so SARIF still omits column keys per the conditional emit policy.

## Architecture

- Domain stays language-agnostic — `SourceSpan` is a value object; columns are universal across syn / tree-sitter / any future tree walker.
- No new dependencies. `proc_macro2` was already pulled in by `syn`.
- Schema version unchanged (additive change with `#[serde(default)]`).

## v0.3.0 milestone state after this merges

5 issues remaining: `#82 → #76 → #77 → #71 → #100`, then cut v0.3.0.

## Summary by CodeRabbit

* **New Features**
  - Column-precise SARIF output: GitHub Code Scanning annotations now underline exact function ranges when column data is available; partial data is omitted to prevent incorrect annotations.

* **Documentation**
  - Updated CHANGELOG with column-precision feature details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 - **`--format sarif`** — emit SARIF v2.1.0 JSON for GitHub Code Scanning. Each function whose CRAP score exceeds the threshold becomes a SARIF `result` with `ruleId: "crap/threshold-exceeded"`, severity mapped from risk level (high → `error`, moderate → `warning`, acceptable & low → `note`), file path + start/end line, and a `partialFingerprints.functionIdentity` for cross-run dedup. SARIF output is a *gate translation*: results derive from the unshapeable analysis, so display flags (`--top`, `--sort-by`, `--only-failing`) do **not** alter SARIF output. `--no-fail` overrides the exit code only — the `results[]` array still reports every finding so PR annotations stay truthful. Pipe stdout into a `.sarif` file and upload via `github/codeql-action/upload-sarif@v3`. Closes #70.
+- **Column-precise SARIF regions** — `domain::types::SourceSpan` gains additive `start_column` / `end_column` fields (1-based; `0` means "unknown"). The Rust complexity adapter populates them from `proc_macro2::Span`, and `--format sarif` emits `region.startColumn` / `endColumn` only when both columns are known. GitHub Code Scanning now underlines the exact function range in the PR diff instead of highlighting the whole line. Adapters without column data (diff hunks) emit `0`, and the reporter omits the column keys to keep half-truths off the wire. JSON envelope `schema_version` is unchanged (additive via `#[serde(default)]`). Closes #105.
 
 ## [0.2.2] - 2026-04-26
 
--- a/src/adapters/complexity/mod.rs
+++ b/src/adapters/complexity/mod.rs
@@ -164,15 +164,25 @@ fn add_contributor(
 }
 
 fn span_of(item: &impl syn::spanned::Spanned) -> SourceSpan {
-    // proc_macro2::LineColumn::line is 1-indexed; column is 0-indexed in
-    // UTF-8 characters. Convert columns to 1-based for SARIF region
-    // semantics. The +1 is intentional and asymmetric with the line copy.
+    // `proc_macro2::LineColumn::line` is 1-indexed already; `column` is
+    // 0-indexed (UTF-8 chars). `Span::start()` is the first character of
+    // the span (inclusive); `Span::end()` is one past the last character
+    // (exclusive). `SourceSpan` stores 1-based inclusive columns, in
+    // parallel with the inclusive `end_line` convention:
+    //
+    //   start_column = start_col_0based + 1
+    //   end_column   = end_col_0based       (because 0-based exclusive
+    //                                        end equals 1-based inclusive
+    //                                        end of the same character)
+    //
+    // Reporters that want SARIF-style exclusive endColumn add 1 at emit
+    // time; everyone else gets the intuitive inclusive bound for free.
     let sp = item.span();
     SourceSpan {
         start_line: sp.start().line,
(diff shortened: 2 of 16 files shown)


cmbays:
Review feedback addressed in 84bb2b2:

| Reviewer | Severity | Item | Disposition |
|----------|----------|------|-------------|
| gemini-code-assist | HIGH | `start_line` missing `#[serde(rename = "startLine")]` | **False positive.** The attribute is present at `sarif.rs:211`, immediately above `start_line: usize` at `:212`. The integration test at `tests/sarif_reporter_integration.rs:115` asserts `region["startLine"]` and has been passing. The bot misread the line. Dismissed. |
| gemini-code-assist | MEDIUM | `ComplexityContributor.column` is 0-based; `SourceSpan` columns (this PR) are 1-based — convention drift | **Valid, out of scope.** Unifying requires a JSON `schema_version` bump (existing breakdown JSON encodes `column: 0` as "first column 0-based"; under 1-based it would mean invalid). Tracked as #107 (`type:chore` `priority:later`). |
| coderabbit | nit | Tighten integration test to catch missing `+1` at the wire boundary | **Applied.** `sarif_region_carries_columns_from_real_source` now asserts `start_col == 1` (FIXTURE_SRC failing fns are unindented) and `end_col > start_col` (single-line span). |
| coderabbit | nit | Use `SourceSpan` import style consistent with `FunctionVerdict` / `RiskLevel` | **Applied.** Lifted `SourceSpan` into the file-level `use` and dropped the fully-qualified path in `region_for_span`. |
| coderabbit | nit | Defensive `.max(1)` on `end_column` against future proc_macro2 fallback returning 0 | **Declined.** A magic clamp masks real issues if proc_macro2 semantics ever shift; the pinned exact-value tests (`columns_are_exact_one_based_offsets_*`) would fail loudly and surface the regression with a useful diff. Defensive insurance against a hypothetical isn't worth the test-debugging cost. |

757/757 tests pass after the fixes. fmt + clippy clean.

PR #129, opened 2026-05-09
fix(action): #128 — version-floor probe for scorecard-row format

### LINKED ISSUE #128: fix(action): row-json output silently empty when --help-grep probe runs against v0.4.0+ CLI
## Summary

The composite `scorecard` action's `outputs.row-json` is silently emitted empty against crap4rs v0.4.0+, even though the installed binary fully supports `--format scorecard-row`. Consumers downstream of the action see a missing/empty artifact and degrade their UI to "producer pending" — masking real CRAP-delta data behind a misleading pending state.

## Acceptance Criteria

- [ ] `outputs.row-json` carries the producer-emitted single-row JSON end-to-end against crap4rs v0.4.0 (and any future release that retains the custom `--format` parser).
- [ ] The graceful-degradation path (older crap4rs without the format) still emits empty + the existing `::warning::`, so the action keeps working when a consumer pins an older `version:` input.

## Context

- **Surfaced by**: [breezy-bays-labs/mokumo#817](https://github.com/breezy-bays-labs/mokumo/pull/817) (closes mokumo#806). mokumo's quality workflow wires the action's `outputs.row-json` into a downstream aggregator artifact; the empty-output regression was visible end-to-end on the first CI run.
- **Related**: #112 (action exposes `outputs.row-json`), #119 (CLI adds `--format scorecard-row`), #100 (CLI adopts multi-format `<format>:<file>` syntax — the change that altered `--help` output shape).
- **Empirical evidence** (mokumo PR #817 commit `6437a52`, run [25415255113](https://github.com/breezy-bays-labs/mokumo/actions/runs/25415255113)):
  - Pre-installed crap4rs detected by the action: `crap4rs 0.4.0 (d88d1d1 2026-05-04)`
  - Action workflow warning: `Installed crap4rs (...) lacks '--format scorecard-row'; outputs.row-json will be empty.`
  - Uploaded artifact size: 0 bytes
  - Direct invocation of the same binary with `--format scorecard-row` on the same runner produces a valid Row JSON.

## Root cause

The action's feature-detection probe at `.github/actions/scorecard/action.yml` runs:

```bash
if crap4rs --help 2>/dev/null | grep -- 'scorecard-row' >/dev/null; then ...
```

This worked when `--format` was a plain `clap::ValueEnum` (clap auto-prints `[possible values: ...]` into `--help`). The multi-format parser introduced in #100 replaces the `ValueEnum` with a custom `FormatSpec` parser. clap can't introspect a custom parser, so `--help` no longer enumerates format names — the literal-string grep finds nothing, and the action's gating logic incorrectly concludes the format is unsupported.

The CLI itself is fine; runtime parsing of `--format scorecard-row` succeeds. The bug is purely in the action's feature-detection.

## Workaround in consumers (until this lands)

mokumo PR #817 bypasses `outputs.row-json` and invokes `crap4rs --format scorecard-row` directly from the consumer workflow with the same coverage / src / baseline / config / threshold inputs the action used. Adds ~10s LCOV reparse per run; reverts cleanly once the probe is fixed upstream.

cmbays:
## Summary

The composite scorecard action's `outputs.row-json` was silently emitted empty against crap4rs v0.4.0+, even though the binary fully supports `--format scorecard-row`. Surfaced by mokumo#817 (crap4rs detected as `0.4.0`, action emitted the `::warning::` and an empty artifact, while a direct invocation of the same binary produced a valid Row JSON).

## Root cause

The probe at `.github/actions/scorecard/action.yml` ran:

```bash
if crap4rs --help 2>/dev/null | grep -- 'scorecard-row' >/dev/null; then ...
```

This worked when `--format` was a plain `clap::ValueEnum` (clap auto-prints `[possible values: ...]`). #100 replaced `--format` with a custom `FormatSpec` parser so `--format` could accept `<format>:<file>` syntax — clap can't introspect a custom parser, so `--help` no longer enumerates format names, and the grep false-negatives against the very versions that support the format.

## Fix

Probe by version, not `--help` text:

- Parse `crap4rs --version` → `crap4rs 0.4.0 (sha date)` → `awk '{print $2}'` → `0.4.0`
- Compare against `0.4.0` floor using `sort -V`
- Empty / unparseable version → falls into the existing graceful-degradation branch (empty row-json + `::warning::`)

`sort -V` is available on both GitHub-hosted Ubuntu and macOS runners, so the same probe works on all action consumers.

## Test plan

- [x] Local probe simulation across `0.3.0`, `0.3.9`, `0.4.0`, `0.4.1`, `0.5.0`, `1.0.0`, empty string — supported/unsupported routing matches expectations.
- [ ] CI run against the dogfood PR binary (this PR's `scorecard.yml`) — `crap4rs 0.4.0` should now hit the supported branch and emit a non-empty `outputs.row-json`.
- [ ] mokumo#817 follow-up: drop the local workaround once this lands and re-check `outputs.row-json` end-to-end.

Closes #128

## Summary by CodeRabbit

* **Chores**
  * Enhanced version detection mechanism for improved compatibility handling. The workflow step now gracefully manages unsupported versions and provides clearer feedback when limitations are encountered.

### DIFF
--- a/.github/actions/scorecard/action.yml
+++ b/.github/actions/scorecard/action.yml
@@ -211,13 +211,23 @@ runs:
         # therefore false-negatives against the very versions that
         # support it. `sort -V` works on both GitHub-hosted Ubuntu and
         # macOS runners.
-        installed_version="$(crap4rs --version 2>/dev/null | awk '{print $2}')"
+        installed_version_raw="$(crap4rs --version 2>/dev/null | awk '{print $2}')"
+        installed_version="${installed_version_raw#v}"
         required_version="0.4.0"
-        if [ -n "$installed_version" ] \
+        # Validate against a strict semver shape before `sort -V`. `sort -V`
+        # treats non-numeric tokens (e.g. "unknown", "garbage") and `v`-prefixed
+        # tags as sorting after pure semver strings, so without this gate they
+        # would false-positive into the supported branch and skip the warning.
+        supports_scorecard_row=false
+        if [[ "$installed_version" =~ ^[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?$ ]] \
           && [ "$(printf '%s\n%s\n' "$required_version" "$installed_version" | sort -V | head -n1)" = "$required_version" ]; then
+          supports_scorecard_row=true
+        fi
+
+        if [ "$supports_scorecard_row" = true ]; then
           crap4rs "${common[@]}" --format scorecard-row --no-fail > "$row_json_file"
         else
-          echo "::warning::Installed crap4rs (${installed_version:-unknown}) is older than $required_version; outputs.row-json will be empty. Pin the action's 'version' input to '$required_version' or later for structured Row::CrapDelta output."
+          echo "::warning::Installed crap4rs (${installed_version:-unknown}) is older than $required_version or reports a non-semver version; outputs.row-json will be empty. Pin the action's 'version' input to '$required_version' or later for structured Row::CrapDelta output."
           : > "$row_json_file"
         fi
 
--- a/.github/actions/scorecard/action.yml
+++ b/.github/actions/scorecard/action.yml
@@ -199,22 +199,25 @@ runs:
 
         # Graceful probe: --format scorecard-row landed in crap4rs 0.4.0
         # (https://github.com/breezy-bays-labs/crap4rs/pull/119). Older
-        # releases — including crates.io's 0.3.0, which `cargo binstall
-        # crap4rs` resolves to until 0.4.0 publishes — don't recognize the
-        # value and clap exits non-zero. Probe the installed help text and
-        # emit an empty row-json with a workflow warning if the variant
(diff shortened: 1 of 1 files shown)


cmbays:
Addressed CodeRabbit's feedback in dd40d99 — gated the version probe behind `^[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z.-]+)?$` so non-semver tokens (`unknown`, `garbage`, `v`-prefixed tags, etc.) route through the graceful-degradation branch with the workflow warning instead of false-positiving into the supported path. Empirically verified across the edge-case table CodeRabbit ran:

| Input | Routes to |
|---|---|
| `` (empty) | FALLBACK |
| `unknown` | FALLBACK |
| `garbage` | FALLBACK |
| `v0.3.9` | FALLBACK |
| `v0.4.0` | SUPPORTED (`v` stripped, then semver-validated) |
| `0.3.0` | FALLBACK |
| `0.4.0` | SUPPORTED |
| `0.4.0-rc.1` | SUPPORTED |

Issue #189, opened 2026-05-13
test(crap4ts): crap4ts@1.x corpus + reference capture for parity oracle


cmbays:
## Summary

Snapshot the crap4ts@1.x TS source tree into `crates/crap4ts/tests/fixtures/crap4ts-v1/`, capture its CRAP scorecard output via one-shot `pnpm run crap` against itself, commit the resulting `crap4ts-v1-reference.json` as the parity oracle for W3.2. Pure data + documentation work — no Rust code changes. Uses `cp -R` (not a submodule — per shaping decision: fixed corpus over moving target).

## Acceptance Criteria

- [ ] `cargo build --workspace` passes (no source changes; fixture is pure test data)
- [ ] `cargo nextest run --workspace --all-targets` passes — no test failures from adding fixtures
- [ ] `git ls-files crates/crap4ts/tests/fixtures/crap4ts-v1/ | wc -l` is the expected count (audit during PR review — no stray `.test.ts` files, no `coverage/` dirs, no `node_modules/`, no `dist/`)
- [ ] Fixture README documents: captured commit hash, Node version, pnpm version, capture commands, tolerance band
- [ ] `crap4ts-v1-reference.json` parses as valid JSON and contains a `summary.functions_analyzed` count consistent with crap4ts@1.x source size
- [ ] Wire snapshots byte-identical (no code changes)

## Context

- Epic: #173
- Pipeline: `crap-rs/20260513-typescript-adapter`
- Plan: `~/Github/ops/workspace/crap-rs/20260513-typescript-adapter/impl-plan.md` — see section **W3.1** for full scope, capture commands, and worktree prompt
- Blocked-by: W2.1 + W2.2 + W2.3 + W2.4 + W2.5 (W2 complete — analyzer must be feature-complete before snapshotting the oracle)
- Shape decision: clone snapshot, not git submodule (fixed corpus)

## Discovery

- Verify `~/Github/crap4ts/` is at a clean commit before capture (record `git rev-parse HEAD`)
- Whether the v1.x reference treats test files as analyzable (decide based on `parity_with_v1.feature` scenarios; include or exclude consistently)
- If `~/Github/crap4ts/` has drifted since the 2026-05-08 parity audit, re-validate fixture capture against the audit-era commit `44eb2c2` (per shaping.md)

PR #216, opened 2026-05-16
fix(crap4ts): #215 — normalize_path suffix-fallback for absolute paths from a different machine

### LINKED ISSUE #215: fix(crap4ts): W3.1 sanity-check reports 0%-coverage on v1.x corpus (path-matching gap)
## Summary

W3.1 (#189) sanity-check exercises `crap4ts@2.x` against the captured `crap4ts@1.x` corpus and reports **0% coverage on all 158 discovered functions** despite the coverage JSON containing valid hit-count data. Root cause is suspected to be a path-matching gap between the walker's discovered file paths (canonical, under the fixture's `--src` tempdir-copy) and the Istanbul coverage records' `path` field (absolute, `/Users/cmbays/Github/crap4ts/src/...` — captured on the originating machine). W2.4's `normalize_path` should bridge this via `strip_prefix(canonical(effective_src))`; apparently doesn't for this specific path shape.

**Blocks #190 (W3.2 parity harness)**: with all coverage at 0%, every function would classify as a "score regression" against the v1.x reference oracle. The three-way classification (threshold-change / regression / improvement) requires meaningful coverage data to work.

## Reproducer

```bash
cd /Users/cmbays/github/crap4rs
./target/release/crap4ts \
  --src crates/crap4ts/tests/fixtures/crap4ts-v1/src \
  --coverage crates/crap4ts/tests/fixtures/crap4ts-v1/coverage-final.json \
  --format json --no-fail \
  | jq '.result.summary | {min_coverage, average_coverage, median_coverage}'
```

Output:
```json
{
  "min_coverage": 0.0,
  "average_coverage": 0.0,
  "median_coverage": 0.0
}
```

Expected: at least some functions should report non-zero coverage (v1.x's reference oracle shows 137 functions with varying coverage percentages; v2.x's walker discovers 158 functions in the same source, so a substantial overlap should match).

## Acceptance Criteria

- [ ] **Diagnosis**: identify whether the failure is in (a) `normalize_path` not stripping the absolute prefix for this path shape, (b) line-range matching not finding statements in the matched files, or (c) somewhere else in the analysis pipeline
- [ ] **Fix**: at least some functions in the v1.x corpus should report non-zero coverage after the fix. The actual percentages are W3.2's domain to classify; this issue closes when the matching pipeline produces *any* coverage signal.
- [ ] **Regression test**: an end-to-end smoke test that locks the v1.x corpus's expected functions-with-coverage count (a deliberate floor, e.g. "≥ 100 of 158 functions report non-zero coverage" — pinning a precise count would be brittle while diagnostic work is in flight; pin a floor instead and tighten in W3.2)
- [ ] **Document**: if the fix changes `normalize_path` semantics, update the W2.4 ADR (D16 — `adr-istanbul-schema-tolerance.md`) to capture the path-shape contract

## Context

Sanity-check details from PR #213 (W3.1 capture):
- Walker discovers **158 functions** in 24 files (v1.x reference oracle: 137 functions; discovery delta is W3.2's classification domain — possibly improvements, possibly walker over-discovery)
- Coverage JSON parses cleanly post-#211 (1,943 `column: null` + empty `branchMap.locations[]` accommodated by `Option<u32>` + dropped field)
- `cargo run --bin crap4ts ... --format json` exits 2 with stderr including a misleading "LCOV parse issue(s)" warning (the warning text appears to be from a shared crap-core code path that doesn't differentiate Istanbul vs LCOV); the diagnostic count is 28
- Coverage shape inspection: the JSON's `path` field is `/Users/cmbays/Github/crap4ts/src/adapters/complexity/facade.ts` (absolute, original-machine path); the fixture's `--src` resolves canonically to `/Users/cmbays/github/crap4rs-issue-189/crates/crap4ts/tests/fixtures/crap4ts-v1/src/`

W2.4's `normalize_path` (`crates/crap4ts/src/adapters/coverage/mod.rs`) is supposed to handle absolute paths via `strip_prefix(canonicalized effective_src)`. The hypothesis: the absolute paths in v1.x's coverage are NOT under the fixture's canonicalized `effective_src`, so `strip_prefix` returns `None`, the records get classified as `PathUnresolved`, and a `PathUnresolved` diagnostic is emitted *per file* (which matches the "28 diagnostics" count — 28 files in the coverage, all unresolvable).

If the diagnosis is correct, the right fix is one of:
- **(a)** Make `normalize_path` smarter: when the Istanbul path is absolute but doesn't share a prefix with `effective_src`, attempt to match by **relative suffix** (e.g. `adapters/complexity/facade.ts` exists under `effective_src` even if the absolute prefix differs)
- **(b)** Rewrite the fixture's coverage JSON paths from `/Users/cmbays/Github/crap4ts/src/...` to relative `src/adapters/...` during capture, so the existing strip-prefix works
- **(c)** Add a CLI flag `--coverage-src-prefix=<prefix>` that the user can pass to manually strip a known absolute prefix (escape hatch)

Strong preference for (a) since it's the more user-friendly path — vitest's Istanbul output for any project will have machine-specific absolute paths that won't match a CI-captured fixture's `--src`. The capture-time rewriting in (b) breaks the "fixture is reproducible from raw producer output" principle.

## Discovery

- Confirm the diagnosis above by `--verbose`-running the analysis and inspecting per-file diagnostics
- Check whether W2.4's `path_unresolved_emits_path_unresolved_diagnostic_and_valid_entries_still_parse` smoke test exercises the same scenario in synthetic form (does it use absolute paths under a different prefix?)
- If suffix-matching becomes the fix, it's worth documenting as an ADR amendment to D16 — the path-mismatch contract gains a "fallback strategy" arm

## Related

- **#189** — W3.1 capture (introduced the corpus that surfaces this gap)
- **#190** — W3.2 parity harness (BLOCKED by this issue; can't classify meaningfully without coverage signal)
- **#211** — closed by PR #213's commit `6d6d29b`; sibling parser-side fix surfaced by the same sanity-check
- **D16** — `adr-istanbul-schema-tolerance.md`; the path-mismatch contract lives here

cmbays:
## Summary

`crap4ts@2.x` run against the captured `crap4ts@1.x` corpus (W3.1, #189) reported **0% coverage on all 158 discovered functions** despite valid hit-count data. Root cause: `IstanbulCoverage::normalize_path` did a single `strip_prefix(effective_src)`. The v1.x corpus's `coverage-final.json` carries machine-absolute paths captured on the originating machine (`/Users/cmbays/github/crap4ts/src/...`); these share no prefix with the fixture's canonicalized `--src`, so `strip_prefix` returned `None` for every entry → 28 `PathUnresolved` diagnostics → zero coverage downstream.

This adds a **suffix-reachability fallback** (issue option (a), the chosen direction): when `strip_prefix` fails, walk the path's components longest→shortest under `effective_src` and accept the first suffix that resolves to a real file.

## Reproducer (before)

```
./target/release/crap4ts --src crates/crap4ts/tests/fixtures/crap4ts-v1/src \
  --coverage crates/crap4ts/tests/fixtures/crap4ts-v1/coverage-final.json \
  --format json --no-fail
```

Before: `min_coverage: 0.0`, `average_coverage: 0.0`, `median_coverage: 0.0`; 158 functions, **0** with non-zero coverage; **28** `PathUnresolved` parse diagnostics; one sample diagnostic path: `/Users/cmbays/github/crap4ts/src/adapters/complexity/facade.ts` (different-machine absolute prefix; the suffix `adapters/complexity/facade.ts` exists under the fixture src).

## Fix mechanism

Two-arm `normalize_path`:
1. **Strip-prefix fast path** (pure, no I/O) — unchanged W2.4 behavior for same-machine captures.
2. **Suffix-reachability fallback** (#215, bounded `.is_file()` I/O) — `suffix_match_under` iterates path-component suffixes longest→shortest, joining each under `effective_src`, returning the first that `.is_file()`-resolves. Longest-suffix-wins is deterministic (disambiguates a leaf filename present in multiple dirs). A `starts_with(effective_src)` guard rejects the degenerate `start==0` case where `effective_src.join(absolute)` collapses back to the absolute path itself (would otherwise leak an out-of-tree match). `.is_file()` chosen over `.exists()` (one syscall, no directory false-positive). `PathUnresolved` still fires as the **final** arm when no suffix matches — the D16 diagnostic-and-skip contract is preserved.

## Acceptance gates checklist

- [x] **1.** `cargo build --workspace` passes
- [x] **2.** `cargo nextest run --workspace --all-targets` — 1031 tests, all green (1027 baseline + 4 new)
- [x] **3.** New smoke tests (a)-(d) pass; (d) enforces the ≥100/158 floor as a hard assertion
- [x] **4.** `cargo fmt --check && cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] **5.** `cargo +1.93 check --workspace --all-targets` passes (MSRV)
- [x] **6.** `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps` clean
- [x] **7.** `cargo deny check` clean (advisories/bans/licenses/sources ok)
- [x] **8.** `lefthook run pre-push` — skipped ("no matching push files": lefthook evaluates against the main repo's hooks-path config, not the worktree diff). Equivalent manual checks (fmt/clippy/test/doc/deny/MSRV) all green.
- [x] **9.** Wire snapshots byte-identical for BOTH crap4rs + crap4ts (`cargo nextest run` both `wire_envelope_*` binaries green, no `.snap.new` files)
- [x] **10.** No drift declared for both wire-envelope modules (see Wire-shape canary below)
- [x] **11.** Plan-of-record deviations subsection present (below)

## Sanity-check findings (before → after)

| Metric | Before | After |
|---|---|---|
| Functions with non-zero coverage | **0 / 158** | **138 / 158** |
| `average_coverage` | 0.0 | 86.35 |
| `median_coverage` | 0.0 | 100.0 |
| Parse diagnostics (`PathUnresolved`) | 28 | 4 |

The 4 remaining `PathUnresolved` are `adapters/coverage/{detect,facade,istanbul,v8}.ts` — files the v1.x corpus referenced in coverage but did **not** capture into the fixture's `src/` tree (28 coverage files vs 24 present in src). The fallback correctly does **not** invent matches for genuinely-absent files — confirms no over-match / residual bug. 138/158 comfortably clears the ≥100/158 floor.

## Plan-of-record deviations

- (i) **`normalize_path` purity invariant deliberately broken** — arm 2 does bounded, OS-cached `.is_file()` filesystem I/O. Required to bridge cross-machine portable fixtures; the pure-strip-prefix arm 1 is preserved as the fast path.
- (ii) **Docstrings updated** — removed the "Pure: no filesystem I/O" and "Mirrors `LcovParser::normalize_path`" claims from both the function docstring **and** the module-level `## Path normalization` section (the latter was an additional stale claim beyond the plan's scope; flagged + fixed).
- (iii) **D16 amended** with the suffix-fallback arm (see cross-reference below).
- (iv) **`.is_file()` chosen over the issue body's `.exists()` suggestion** — defensive tightening (one syscall; prevents a directory false-positive when a raw path component matches a directory name under root).
- (v) pure-with-injected-file-list alternative considered; rejected — requires constructor signature change beyond #215 scope.
- (vi) **lowercase-g calibration** — plan stated `/Users/cmbays/Github/...` (capital G); fixture actually carries `/Users/cmbays/github/...` (lowercase). Mechanism (prefix mismatch — different project tree, not case-sensitivity) unchanged; recorded for future-reader accuracy. No diagnosis deviation.
- **Added defensive guard beyond the plan's prescribed `normalize_path`**: a `starts_with(effective_src)` check in `suffix_match_under` rejecting the `start==0` degenerate case (absolute RHS collapses `Path::join`). Surfaced by advisor pre-implementation review; without it an absolute path existing verbatim outside `effective_src` would leak an out-of-tree match. Documented in the `suffix_match_under` docstring; exercised by test (c).
- **No existing W2.4 test changed behavior** — `w24_nyc_absolute_paths_normalize_via_strip_prefix` and `w24_orphan_path_emits_path_unresolved_and_valid_entries_still_parse` both still pass unchanged (verified). No test changes outside `istanbul_smoke.rs`.

## Wire-shape canary: no drift

`normalize_path` is parser-internal and does not surface in the JSON envelope. Both `wire_envelope_crap4rs` and `wire_envelope_crap4ts` snapshot tests pass byte-identical; no `.snap.new` files generated. **No drift for either wire-envelope module.**

## D16 amendment cross-reference

D16 (`adr-istanbul-schema-tolerance.md`) amended with a new "Suffix-fallback arm (#215)" subsection under "Path-mismatch contract", plus `#215` added to References and the "Out of scope" filesystem-canonicalization bullet updated. **This amendment lives in the separate `ops` repo (`~/Github/ops/decisions/crap-rs/adr-istanbul-schema-tolerance.md`) and is NOT part of this PR's diff.** The D16 subsection was further amended for the #216 traversal guard (see Security follow-up below).

## Security follow-up (gemini #216)

Commit `d77ede9` addresses a **gemini-code-assist security-medium** finding on `normalize_path`.

**Finding + scope expansion (honest reporting):** gemini flagged that `Path::starts_with` is lexical, so a `..`-containing suffix could pass the under-root guard in `suffix_match_under` (arm 2) yet `.is_file()`-resolve *outside* `effective_src`. A standalone repro proved **arm 1 (`strip_prefix`) has the identical defect** — it is also lexical:

```
effective_src = /tmp/x/project
raw           = /tmp/x/project/../outside/secret.ts
strip_prefix:            Ok("../outside/secret.ts")   ← succeeds, points OUT of tree
joined.starts_with(eff): true                          ← lexical, also fooled
```

Since `coverage-final.json` is **user-supplied**, a crafted `..`-traversal path could probe / bind coverage to out-of-tree files via *either* arm. The fix is therefore an **authoritative ParentDir-rejection guard at `normalize_path`'s return point, covering both arms uniformly** (broader than gemini's stated arm-2-only finding). `suffix_match_under` retains a per-iteration `..` skip as a local perf win (avoids a doomed `.is_file()` syscall) + defense-in-depth.

**New regression test (e):** `fix216_parentdir_traversal_does_not_resolve_outside_effective_src` — constructs `<src_root>/../outside/secret.ts` against a genuinely-existing on-disk sibling sentinel, asserts exactly one `PathUnresolved` and that `secret.ts` never enters `out.coverage`, no panic.

**Floor unchanged:** the #215 (d) e2e floor is still **138/158** post-`..`-filter — the traversal guard cost **zero** legitimate corpus matches (v1.x corpus paths are flat, no legitimate `..`).

**Re-verified gates post-`d77ede9`:** build / nextest --workspace --all-targets (1032 tests, all green) / clippy -D warnings / fmt --check / cargo +1.93 MSRV / doc -D warnings / cargo deny / wire-envelope byte-identical (no `.snap.new`). D16 amended (separate ops repo, not in this diff) with the traversal-guard sentence.

Closes #215

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of Istanbul coverage files that contain absolute paths from different machines through enhanced path resolution and fallback matching.

* **Tests**
  * Added regression test suite for cross-machine coverage file path resolution ensuring robust portable path handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/breezy-bays-labs/crap-rs/pull/216?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/crap4ts/src/adapters/coverage/mod.rs
+++ b/crates/crap4ts/src/adapters/coverage/mod.rs
@@ -43,14 +43,16 @@
 //!
 //! ## Path normalization
 //!
-//! Mirrors `crap4rs::adapters::coverage::LcovParser::normalize_path` —
-//! a pure `strip_prefix(effective_src)`. The orchestrator pre-
-//! canonicalizes `effective_src` at the factory-closure boundary, so
-//! the parser does no filesystem I/O during path joining. Entries
-//! whose paths fall outside `effective_src` emit an
-//! `IstanbulParseDiagnostic { kind: PathUnresolved, … }` and are
-//! skipped — per Resolved Q10 in shaping, the scorecard NEVER aborts
-//! on first failure and NEVER silent-drops a record.
+//! Two-arm strategy (see `IstanbulCoverage::normalize_path`): a pure
+//! `strip_prefix(effective_src)` fast path for same-machine captures,
+//! then a bounded `.is_file()` longest-suffix reachability fallback
+//! (#215) for portable fixtures whose absolute paths were captured on
+//! a different machine and share no prefix with the local
+//! `effective_src`. The orchestrator pre-canonicalizes `effective_src`
+//! at the factory-closure boundary. Entries whose paths resolve under
+//! neither arm emit an `IstanbulParseDiagnostic { kind: PathUnresolved,
+//! … }` and are skipped — per Resolved Q10 in shaping, the scorecard
+//! NEVER aborts on first failure and NEVER silent-drops a record.
 
 use crap_core::domain::types::{BranchCoverage, CrapError, LineCoverage};
 use crap_core::ports::{CoveragePort, ParseOutput};
@@ -90,15 +92,41 @@ impl IstanbulCoverage {
         }
     }
 
-    /// Strip `effective_src` from `raw` and return a workspace-relative
-    /// path, or `None` if the entry resolves outside the source tree.
+    /// Resolve `raw` to a workspace-relative path under
+    /// `effective_src`, or `None` if it can't be resolved.
     ///
-    /// Pure: no filesystem I/O. Mirrors `LcovParser::normalize_path`
-    /// in `crap4rs`. The orchestrator canonicalizes `effective_src`
-    /// before construction, so no further `canonicalize()` walk is
-    /// needed here. Joining relative entries against `effective_src`
-    /// and stripping the prefix is the same two-step the LCOV parser
-    /// uses on its own platform.
+    /// Two-arm strategy:
+    ///
+    /// 1. **Strip-prefix fast path (pure, no I/O).** Join relative
+    ///    entries against `effective_src`, then strip the canonical
+    ///    `effective_src` prefix. This succeeds whenever the coverage
+    ///    payload was captured on the same machine/tree being analyzed
+    ///    (jest/vitest/nyc on the local checkout). The orchestrator
+    ///    pre-canonicalizes `effective_src`, so no `canonicalize()`
+    ///    walk is needed here.
+    /// 2. **Suffix-reachability fallback (bounded `.is_file()` I/O,
+    ///    #215).** When the coverage was captured on a *different*
+    ///    machine (a portable fixture: coverage produced on machine A,
+    ///    analyzed on machine B), the entry's absolute path shares no
+    ///    prefix with the local `effective_src` and arm 1 returns
+    ///    `None`. Arm 2 then walks the path's components longest →
+    ///    shortest, joining each suffix under `effective_src` and
+    ///    accepting the first suffix that resolves to a real file. The
+    ///    longest matching suffix wins, which deterministically
+    ///    disambiguates a leaf filename (e.g. `index.ts`) that exists
+    ///    in multiple directories — the longer shared path is the
+    ///    structural truth; the machine-specific absolute prefix is
+    ///    noise. If two equal-length suffixes could both match (only
+    ///    reachable via symlinks/odd trees), the first found wins.
+    ///
+    /// This is *not* a pure function and no longer mirrors
+    /// `crap4rs::adapters::coverage::LcovParser::normalize_path` (which
+    /// remains pure strip-prefix): the fallback does bounded,
+    /// OS-cached filesystem I/O (~components × `.is_file()` syscalls).
+    /// That trade buys cross-machine fixture portability. When even the
+    /// suffix match fails, this returns `None` and the caller emits
+    /// `PathUnresolved` as before — the diagnostic-and-skip contract
+    /// (D16) is preserved as the final arm.
     fn normalize_path(&self, raw: &str) -> Option<PathBuf> {
         let path = PathBuf::from(raw);
         let joined = if path.is_absolute() {
@@ -106,10 +134,46 @@ impl IstanbulCoverage {
         } else {
             self.effective_src.join(raw)
         };
-        joined
-            .strip_prefix(&self.effective_src)
-            .ok()
-            .map(|p| p.to_path_buf())
+        // Arm 1: strip the canonical effective_src prefix (W2.4 fast
+        // path — same-machine capture).
+        if let Ok(stripped) = joined.strip_prefix(&self.effective_src) {
+            return Some(stripped.to_path_buf());
+        }
+        // Arm 2: cross-machine absolute paths don't share a prefix —
+        // find the longest path suffix that resolves to a real file
+        // under effective_src (#215).
+        self.suffix_match_under(&joined)
+    }
+
+    /// Find the longest suffix of `raw` whose components, joined under
+    /// `effective_src`, resolve to a real file. Returns the
+    /// workspace-relative suffix, or `None` if no suffix is reachable.
+    ///
+    /// Iterates suffixes longest → shortest (`start` from 0 outward),
+    /// so the first reachable suffix is the longest one
+    /// (longest-suffix-wins precedence — see [`Self::normalize_path`]).
+    ///
+    /// `.is_file()` is used (not `.exists()`): one syscall, and it
+    /// prevents a directory false-positive when a raw path component
+    /// happens to match a directory name under `effective_src`.
+    ///
+    /// The `starts_with(effective_src)` guard rejects the `start == 0`
+    /// degenerate case: when `raw` is absolute, `effective_src.join(raw)`
+    /// collapses back to `raw` itself (Rust `Path::join` replaces the
+    /// base with an absolute RHS). Without the guard, an absolute path
+    /// that exists verbatim *outside* `effective_src` would leak an
+    /// out-of-tree match. The guard keeps every accepted match strictly
+    /// under the source root.
+    fn suffix_match_under(&self, raw: &Path) -> Option<PathBuf> {
+        let components: Vec<_> = raw.components().collect();
+        for start in 0..components.len() {
+            let candidate_rel: PathBuf = components[start..].iter().collect();
+            let candidate_abs = self.effective_src.join(&candidate_rel);
+            if candidate_abs.starts_with(&self.effective_src) && candidate_abs.is_file() {
+                return Some(candidate_rel);
+            }
+        }
+        None
     }
 }
 
--- a/crates/crap4ts/src/adapters/coverage/mod.rs
+++ b/crates/crap4ts/src/adapters/coverage/mod.rs
@@ -127,6 +127,22 @@ impl IstanbulCoverage {
     /// suffix match fails, this returns `None` and the caller emits
     /// `PathUnresolved` as before — the diagnostic-and-skip contract
     /// (D16) is preserved as the final arm.
+    ///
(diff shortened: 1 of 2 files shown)


cmbays:
@gemini-code-assist Addressed in commit d77ede9.

Your security-medium finding (lexical `Path::starts_with` bypassed by `..` in the suffix fallback) was correct. A standalone repro showed the **strip-prefix fast path (arm 1) has the identical lexical-vs-resolved defect** — `strip_prefix("/root/project", "/root/project/../outside/x.ts")` returns `Ok("../outside/x.ts")`, an out-of-tree relative path. So the fix is an **authoritative `Component::ParentDir` rejection at `normalize_path`'s return point, covering both arms uniformly** (broader than the stated arm-2-only finding). `suffix_match_under` keeps a per-iteration `..` skip for local perf + defense-in-depth.

Regression test `fix216_parentdir_traversal_does_not_resolve_outside_effective_src` constructs a `<src_root>/../outside/secret.ts` traversal against a genuinely-existing on-disk sibling and asserts `PathUnresolved` + the sentinel never enters `out.coverage`. The #215 corpus floor is unchanged at 138/158 (the `..` filter cost zero legitimate matches). All 11 acceptance gates re-verified green. Resolving this thread.

PR #257, opened 2026-05-23
fix(crap4ts): #252 — collapse multi-statement-per-line via MIN (arrow undercount fix)

### LINKED ISSUE #252: fix(crap4ts): never-invoked single-line arrow reports ~50% coverage instead of 0%
## Summary

`crap4ts` reports **~50% line coverage for a never-invoked single-line arrow function**, not 0%. When an arrow is declared on one line — `export const cube = (x: number) => x * x * x;` — Istanbul records two statements on that line: the `const` declaration (runs at module load → hit) and the arrow body (never invoked → 0 hits). crap4ts's per-function line-coverage rollup counts the line as partially covered, so `cube`'s function coverage comes out ~50% even though its body never ran.

The W1.1 arrow-coverage smoke tests (`istanbul_smoke.rs::arrow_ac_5a`) work around this by asserting at the **statement** level ("at least one statement at line 2 has 0 hits") rather than asserting `cube`'s function coverage is 0.0 — because it isn't.

## Acceptance Criteria

- [ ] A single-line arrow whose body is never invoked reports 0% (not ~50%) function-level line coverage — the `const`/`let` declaration statement on the same line is excluded from the function's own span coverage
- [ ] `square` (invoked) still reports 100%; the fix does not regress covered arrows
- [ ] `arrow_function_coverage.feature` scenario 1 ("An invoked arrow function has matching coverage") is wired `@unwired` → `@wired` against the corrected behavior

## Context

- Epic: #173
- Surfaced during #229 (crap4ts BDD wiring). `arrow_function_coverage.feature` scenario 1 asserts `cube` line coverage `0.0` and explicitly guards against "silent undercount"; crap4ts@2 currently produces `50.0`, so the scenario cannot be wired truthfully. #229 wires scenarios 2–4 and leaves scenario 1 `@unwired`, tracked here.
- This is one of the three documented compounding score-divergence reasons in `MIGRATION.md` ("possible arrow-function undercount").
- `priority:later` — not GA-blocking; the per-statement data is correct, only the per-function rollup conflates the declaration line.

## Discovery

- Whether the fix belongs in the crap4ts line-range matcher (exclude the declaration line from the arrow's own span) or in `crap-core::domain::matching` (the line-range join).
- Whether multi-line arrows are affected (likely not — the declaration and body are on different lines).
- Whether the same conflation affects `const x = function() {…}` function expressions.

cmbays:
## Summary

Pre-fix, `IstanbulCoverage::line_coverage_for` emitted one `LineCoverage` record per Istanbul statement. A **single-line arrow whose body shares its `const` declaration's source line** (`export const cube = (x) => x*x*x;`) produced two records at that line — one for the declaration (hits = 1 at module load) and one for the body (hits = 0 if uninvoked). The matcher's per-function rollup then read `total=2, covered=1, percent=50%` for an arrow that never ran.

This change collapses multi-statement-per-line via `min(hits)` at the parser, aligning the Istanbul adapter with the implicit per-line contract LCOV already obeys (one `DA:line,hits` per line via its `BTreeMap` dedupe).

Cube now reports **0%** function-level line coverage (down from ~50%); square still reports 100%. The MIN semantic is conservative: a line counts as covered iff every statement on it ran ≥ 1×.

Closes #252

## Empirical orientation (advisor-driven; verified before fixing)

- **LCOV contract is implicit "one record per line per file"** — `crap4rs/src/adapters/coverage/mod.rs` populates a `BTreeMap<line, hits>` per `SF:` block, so each line has exactly one record. Confirms the fix layer is the Istanbul adapter (the contract violator), not the language-agnostic matcher.
- **Walker emits cube's span as `(start_line=2, end_line=2)`** — single line — verified from the existing wire-envelope snapshot. The conflation is purely about multi-statement-per-line, not walker drift.
- **Bug scope (precise):** single-line arrows / function expressions where the declaration and body share a source line. **Multi-line arrows are NOT addressed by this PR** — whether they exhibit the same 50% reading depends on the walker's span emission (does it include the declaration line in the body's span?), which is a separate question worth verifying in a follow-up.

## Acceptance criteria (from #252 body)

- [x] A single-line arrow whose body is never invoked reports **0%** function-level line coverage (not ~50%)
- [x] `square` (invoked) still reports **100%**; no regression on covered arrows
- [x] `arrow_function_coverage.feature` scenario 1 flipped `@unwired` → `@wired` in this PR

## What changed

| File | Change |
|---|---|
| `crates/crap4ts/src/adapters/coverage/mod.rs` | `line_coverage_for` aggregates per-line via MIN; 5 new focused unit tests pin the rule |
| `crates/crap4ts/tests/istanbul_smoke.rs` | 5a now asserts one record per line + line 2 hits=0; 5b/5c hits assertions adjusted for MIN semantics; `hits_at` docstring updated |
| `crates/crap4ts/tests/features/arrow_function_coverage.feature` | scenario 1 `@unwired` → `@wired`; tracking comment replaced with resolution comment |
| `crates/crap4ts/tests/arrow_function_coverage_cucumber.rs` | Added `then_not_named_function_coverage` step def to wire the "does NOT show silent undercount" guard step; comments updated |
| `crates/crap4ts/tests/parity_helpers/mod.rs` | New `Crap252Improvement` classification (bidirectional); see "Parity oracle expansion" below |
| `crates/crap4ts/tests/parity_v1.rs` | 3 new tests pinning the bidirectional classifier + CC-drift exclusion |
| `crates/crap-core/tests/snapshots/wire_envelope_crap4ts__envelope.snap` | Drift accepted (cube row + summary stats) |
| `MIGRATION.md` | Bullet 3 updated — the fix **narrows** the documented v1-vs-v2 divergence rather than introducing one |

## Wire envelope drift (one canary, documented-accept)

Only `wire_envelope_crap4ts` drifts — and surgically: only the `cube` row and the summary's `min_coverage` shift. All other functions in the canary are byte-identical.

| Field | Before | After |
|---|---|---|
| `cube.coverage_percent` | 50.0 | 0.0 |
| `cube.crap.value` | 1.13 | 2.0 |
| `summary.min_coverage` (and `worst_function` mirror) | 50.0 | 0.0 |

The other two canaries — `wire_envelope_crap4rs` (LCOV path) and `wire_envelope_crap4ts_branches` (branch-heavy.ts fixture, no shared-line arrows) — stay **byte-identical**.

## Parity oracle expansion — `Crap252Improvement` (bidirectional)

The first run after the fix surfaced **30+ functions** in the v1.x corpus that the existing classifier would have marked as `ScoreRegression`. These are NOT regressions — they are the same #252 mechanism playing out against the v1.x source, just like #37 already classifies the v1 0%-coverage bug as an improvement.

`Crap252Improvement` triggers on two directions:

1. **Coverage deflation** (cube-style): `same_cc && v2.coverage < v1.coverage - COV_EPS && v2.crap >= v1.crap - CRAP_EPS`. An uninvoked body shared its declaration's line; v1 counted the line as half-covered, v2's MIN catches the 0-hit signal.
2. **Coverage inflation** (cleanup): `same_cc && v2.coverage > v1.coverage + COV_EPS && v2.crap <= v1.crap + CRAP_EPS`. A function's span contained lines with multiple duplicate uncovered statements; v1's denominator was inflated by phantom records, v2's MIN collapses them and the function's true coverage % rises.

The `same_cc` guard is load-bearing: it prevents walker-drift cases (different CC pre/post) from being swallowed under the #252 banner. CC mismatches still fall through to `ScoreRegression` and require their own disposition. Verified by the new `crap252_improvement_does_not_swallow_cc_drift` test.

Post-fix parity status:

```
parity gate: PASS
exact-CC rate: ≥ 95%
unexplained regressions: 0
```

Note on `adapters/coverage/istanbul.ts::IstanbulCoverageAdapter.parse` (a function I initially flagged as a possible walker-CC drift): CC matches between v1 and v2; the divergence was purely #252-direction-1 and absorbs cleanly under the new classifier.

## Gate battery

| Gate | Status |
|---|---|
| `cargo fmt --check` | ✅ |
| `cargo clippy --workspace --all-targets -- -D warnings` | ✅ |
| `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps` | ✅ |
| `cargo +1.93 check --workspace --all-targets` | ✅ |
| `cargo deny check` | ✅ |
| AST-purity grep (`crates/crap-core/src/`) | ✅ |
| `cargo nextest run --workspace --all-targets` | ✅ (1105/1105) |
| All 8 cucumber harnesses (`json_reporter`, `metric_unsupported`, `cyclomatic_walker`, `parity_with_v1`, `istanbul_parser`, `arrow_function_coverage`, `file_extensions`, `branch_coverage`) | ✅ (71/71 scenarios) |
| Self-CRAP-thrice (crap-core 953 fns / crap4rs 202 fns / crap4ts 97 fns, strict-15) | ✅ (worst 13.0 / 11.0 / 12.0) |

## Discovery answers (from #252 body)

- **Where does the fix belong** — the crap4ts Istanbul adapter (`line_coverage_for`). The matcher's per-line contract was already implied by the LCOV adapter; the Istanbul parser was the contract violator. Adapter-level fix is surgical and doesn't pollute crap-core's matcher contract.
- **Multi-line arrows** — **NOT addressed by this PR**. Multi-line arrows have the declaration on line 1 and body on line 2+ (different lines, so MIN never aggregates them). Whether they still exhibit a 50% reading depends on the walker's span emission, which is a separate concern. Worth a follow-up fixture to lock the answer either way.
- **Function expressions** (`const x = function() {...}`) — single-line function expressions exhibit the same conflation by the same mechanism, and the fix addresses them by the same MIN aggregation. Multi-line function expressions are in the same boat as multi-line arrows (separate from this PR's scope).

## Follow-ups

- A small fixture verifying multi-line arrow + multi-line function-expression behavior post-fix (does the 50% reading still appear? if so, the next fix is at the walker's span emission, not the matcher).

## Release Notes

* **Bug Fixes**
  * Fixed arrow function coverage reporting to accurately reflect actual coverage instead of underestimated values
  * Improved coverage calculation when multiple statements exist on the same source line

* **Documentation**
  * Updated migration guide explaining new coverage behavior and calculation improvements

* **Tests**
  * Enhanced test validation for updated coverage calculation logic

[Review Change Stack](https://app.coderabbit.ai/change-stack/breezy-bays-labs/crap-rs/pull/257?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -51,22 +51,24 @@ self-diagnose rather than chase a phantom regression:
 
 **Resolved in 2.0.0 — arrow-function coverage now reads correctly.**
 v2's Istanbul adapter consumes `s` / `statementMap` (not `f` / `fnMap`,
-which has known variance across emitters). The 2.0.0-rc.1 / rc.2
-releases emitted one `LineCoverage` record per Istanbul statement, so
-a single source line carrying both a `const` declaration and its arrow
-body produced two records — and the matcher's per-function rollup
-treated a never-invoked arrow as ~50% covered (the declaration's
-module-load hit dominated the body's zero-hit signal). `2.0.0`
-collapses multi-statement-per-line via `min(hits)`, matching the
-implicit per-line contract the LCOV adapter already obeys. The
+which has known variance across emitters). Earlier 2.0.0 release
+candidates emitted one `LineCoverage` record per Istanbul statement,
+so a single source line carrying both a `const` declaration and its
+arrow body produced two records — and the matcher's per-function
+rollup treated a never-invoked arrow as ~50% covered (the
+declaration's module-load hit dominated the body's zero-hit signal).
+`2.0.0` collapses multi-statement-per-line via `min(hits)`, matching
+the implicit per-line contract the LCOV adapter already obeys. The
 practical effect across the `crap4ts@1.x` parity corpus is that **v2
 now reports more accurate coverage than v1 did for any function whose
 body shares a source line with its declaration** — including
 single-line arrows, single-line function expressions, and inline
-`xs.map(arrow)` patterns. Both directions of v1-vs-v2 movement (v2
-deflating an undercount-masked uninvoked body, v2 inflating away
-phantom duplicate-uncovered statements) classify as improvements
-in the parity harness. Tracked: [`crap-rs`#252](https://github.com/breezy-bays-labs/crap-rs/issues/252).
+`xs.map(arrow)` patterns. Both directions of v1-vs-v2 movement
+classify as improvements in the parity harness: v2 reports **lower**
+coverage when a genuinely uninvoked body was previously masked by its
+declaration's module-load hit, and v2 reports **higher** coverage
+when phantom duplicate per-line records previously inflated the
+denominator. Tracked: [`crap-rs`#252](https://github.com/breezy-bays-labs/crap-rs/issues/252).
 
 ### Subpath export removals
 
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -33,9 +33,9 @@ Windows and Linux arm64 / musl are tracked for a later release.
 ### Why are my scores different?
 
 `crap4ts@2.x` scores can diverge from `crap4ts@1.x` scores on the same
-codebase. Three compounding reasons account for the gap; each is
-documented so you can self-diagnose rather than chase a phantom
-regression:
+codebase. Two compounding reasons account for the gap (a third was
+resolved in 2.0.0 — see below); each is documented so you can
+self-diagnose rather than chase a phantom regression:
 
 1. **The default threshold changed from `12` to `16`** (cyclomatic). This
    is an intentional calibration introduced with `crap-rs` v0.5.0 —
@@ -48,18 +48,25 @@ regression:
    against the `crap4ts@1.x` test corpus; until that completes, treat
    the gate as a guideline rather than a ground truth. Track at
    [`crap-rs`#173](https://github.com/breezy-bays-labs/crap-rs/issues/173).
-3. **Arrow-function coverage may count differently.** v1 derived
-   function-level coverage from Istanbul's `f` / `fnMap` arms (per-
-   function invocation counts). v2 reads `s` / `statementMap` only —
-   line-level statement coverage. The two paths agree for the common
-   case where every function's body contains at least one statement
-   counted in `s`, but they can diverge for arrow-heavy code that hits
-   statements without invoking the surrounding function, or vice
-   versa. If a single function's CRAP looks surprisingly different
-   between v1 and v2, eyeball its coverage and file an issue with the
-   fixture; the deliberate trade-off in `crap4ts@2.x` is that
-   modelling `f` / `fnMap` would re-introduce a whole-file bail
-   vector that the v2 parser explicitly avoids.
+
+**Resolved in 2.0.0 — arrow-function coverage now reads correctly.**
+v2's Istanbul adapter consumes `s` / `statementMap` (not `f` / `fnMap`,
+which has known variance across emitters). The 2.0.0-rc.1 / rc.2
+releases emitted one `LineCoverage` record per Istanbul statement, so
+a single source line carrying both a `const` declaration and its arrow
+body produced two records — and the matcher's per-function rollup
+treated a never-invoked arrow as ~50% covered (the declaration's
+module-load hit dominated the body's zero-hit signal). `2.0.0`
+collapses multi-statement-per-line via `min(hits)`, matching the
+implicit per-line contract the LCOV adapter already obeys. The
+practical effect across the `crap4ts@1.x` parity corpus is that **v2
+now reports more accurate coverage than v1 did for any function whose
+body shares a source line with its declaration** — including
+single-line arrows, single-line function expressions, and inline
+`xs.map(arrow)` patterns. Both directions of v1-vs-v2 movement (v2
+deflating an undercount-masked uninvoked body, v2 inflating away
+phantom duplicate-uncovered statements) classify as improvements
+in the parity harness. Tracked: [`crap-rs`#252](https://github.com/breezy-bays-labs/crap-rs/issues/252).
 
 ### Subpath export removals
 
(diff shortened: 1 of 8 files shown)


cmbays:
## Bot review addressed — finding → fix map

Pushed `aeea27d` covering all four findings from CodeRabbit + gemini.

| # | Reviewer | Location | Finding | Fix |
|---|---|---|---|---|
| 1 | gemini-code-assist | `crates/crap4ts/src/adapters/coverage/mod.rs:453` | Pre-allocate `HashMap` with capacity hint based on `entry.s.len()` to avoid rehashing during grouping | `HashMap::with_capacity(entry.s.len())` — at most one per-line record per Istanbul statement; MIN aggregation only shrinks the map, so `entry.s.len()` is a safe upper bound |
| 2 | CodeRabbit | `crates/crap4ts/tests/parity_v1.rs::crap252_improvement_does_not_swallow_cc_drift` | Add same-CC negative case where coverage drifts but CRAP moves the wrong way > `CRAP_EPS`, asserting `ScoreRegression` | Added `inconsistent_direction` case: same CC, coverage 80 → 40 (DOWN), CRAP 2.5 → 1.5 (also DOWN — violates the monotone-decreasing-CRAP-in-coverage invariant at fixed CC, so cannot be the #252 mechanism). Asserts `ScoreRegression` + `!gate_passes()` |
| 3 | CodeRabbit | `MIGRATION.md:52-54` | "2.0.0-rc.1 / rc.2 releases emitted..." may confuse if rc.2 isn't published | Reworded to "Earlier 2.0.0 release candidates emitted..." — decouples the prose from any specific RC count |
| 4 | CodeRabbit | `MIGRATION.md:66-69` | "v2 inflating away phantom duplicate-uncovered statements" is hard to parse | Replaced the parenthetical with explicit lower/higher-coverage cause-and-effect framing |

Verification:
- `cargo nextest run --package crap4ts --test parity_v1` — 13/13 pass (new `inconsistent_direction` case green)
- `cargo nextest run --package crap4ts --test istanbul_smoke` — 32/32 pass (capacity hint is behaviorally invisible)
- `cargo fmt --check && cargo clippy --workspace --all-targets -- -D warnings` — green

Issue #161, opened 2026-05-11
AdapterMeta polish: clap string feature, lifetime tightening, debug_assert guards, test gaps


cmbays:
## Summary

Follow-up from PR C (#148+#152+#160) — review surfaced several polish opportunities for the newly-introduced \`AdapterMeta<'a>\` struct that weren't critical for v0.5.

## Acceptance Criteria

- [ ] **Enable clap \`string\` feature**: add \`features = ["derive", "string"]\` to workspace clap dep so \`Str: From<String>\` exists. Delete the four \`Box::leak\` calls in \`build_command\` + the doc comment explaining the workaround. Net: −12 LOC.
- [ ] **Tighten lifetime on \`extensions\`**: change \`extensions: &'a [&'a str]\` → \`extensions: &'a [&'static str]\` to model actual usage (binaries supply \`const &[&str]\` literals).
- [ ] **Consider further: \`&'static\` on outer lifetime**: drop \`<'a>\` parameter entirely since no test fixture or production caller needs non-static metas. Removes the generic from every public signature.
- [ ] **debug_assert guards**: in \`parse_args\` or \`AdapterMeta::new\` (if introduced), assert \`!config_file_name.is_empty()\`, \`!tool_name.is_empty()\`, \`!tool_version.is_empty()\`. Fires in tests, zero-cost in release.
- [ ] **Adapter-name CI grep gate**: add a sibling check in \`.github/workflows/ci.yml\` ast-purity job that fails on \`"crap4rs"\` / \`"crap4ts"\` string literals (not doc comments) in \`crates/crap-core/src/\`. Prevents silent re-coupling.
- [ ] **\`AdapterMeta\` unit tests**: direct construction + \`extensions_owned()\` round-trip. No test currently constructs it (only main.rs files do).
- [ ] **\`discover_config(name)\` parameterized test**: pass a non-\`crap4rs.toml\` name to verify the lookup is actually param-driven.
- [ ] **\`--help\` content assertion**: integration test asserting \`crap4rs --help\` stdout contains the \`ABOUT\` and \`LONG_ABOUT\` strings.
- [ ] **\`format_extensions_list\` dedup**: lift the duplicated extension-list-formatting match block out of \`core::ensure_source_files_found\` and \`cli::format_extensions_list\` into a single shared \`pub(crate)\` helper.
[...]

Issue #145, opened 2026-05-09
ci(ast-purity): harden grep against grouped imports + add dep-graph fallback


cmbays:
## Summary

The `ast-purity` CI grep introduced in #133 catches the common `use syn::Foo;` pattern but has known blind spots that a determined or accidental import can slip through.

## Known gaps (Gemini Code Assist, PR #144)

The current regex:

\`\`\`
^[[:space:]]*(use|extern[[:space:]]+crate)[[:space:]]+(::)?(syn|quote|proc_macro2?|tree_sitter|swc|oxc)([_:; ]|$)
\`\`\`

Misses these legitimate Rust forms:

- **Grouped imports**: \`use { syn, quote };\` or \`use { syn::Foo, syn::Bar };\` (the regex requires the crate name to be the first token after \`use\`)
- **Multi-line nested groups**: \`use syn::{\\n    parse_macro_input,\\n    Error,\\n};\` is matched by the current regex (since \`syn\` IS first), but \`use {\\n    syn::Foo,\\n};\` is NOT matched.
- **Path-prefix imports**: any future case where the AST crate appears in a non-leading position.

These are edge cases — code review (humans + AI assistants like CR/Gemini) catches most — but the regex's job is to be a **canary that fails fast at CI time**, not just review-time.

## Acceptance criteria

- [ ] The CI \`ast-purity\` job rejects the grouped-import form: \`use { syn };\` should fail the gate.
- [ ] The CI \`ast-purity\` job rejects nested-group AST imports: \`use { syn::Foo };\`.
- [ ] False-positive baseline holds: \`crates/crap-core/src/\` continues to pass on its current source content.
- [ ] Failure messages tell the human reviewer the crate name + line number, not just \`exit 1\`.

## Context

The plan-phase shaping doc explicitly chose \`grep\` over \`cargo-deny\` transitive bans because \`serde_derive\` (a benign proc-macro) would false-positive on a transitive ban. See \`~/Github/ops/workspace/crap-rs/20260509-monorepo-migration/shaping.md\` "Patterns NOT copied" #3.

Hardening options (pick one or stack):

1. **Augment the regex** — add a second pass for \`use[[:space:]]*\\{[^}]*\\b(syn|quote|...)\\b\` patterns. Covers single-line group imports.
2. **\`syn\`-based linter** — a tiny Rust crate that walks \`crates/crap-core/src/\` AST. Most robust, highest maintenance.
[...]

PR #286, opened 2026-05-24
chore(ci): #261 — ast-purity layer-4 adapter-vocabulary string-literal ban

### LINKED ISSUE #261: chore: ast-purity layer-4 — ban adapter-vocabulary string literals in crap-core/src/
## Summary

Extend the `ast-purity` CI gate at `.github/workflows/ci.yml:442` with a **layer-4 string-literal check** that bans adapter-vocabulary literals (`"LCOV"`, `"Istanbul"`, `"TypeScript"`, etc.) in `crates/crap-core/src/`. Closes the same anti-pattern class as #161 + the variant rename in #178 — adapter names should not appear in the language-agnostic domain layer except in legitimate cross-reference rustdoc.

## Acceptance Criteria

- [ ] CI gate step rejects new `#[error("…")]` attribute literals containing `LCOV`, `Istanbul`, `TypeScript`, `LCOV-flavoured` (case-insensitive token match) inside `crates/crap-core/src/`.
- [ ] CI gate rejects user-facing string literals (e.g. `format!("Failed to parse LCOV…")`) carrying the same tokens.
- [ ] CI gate **does not** false-positive on:
  - rustdoc `///` lines (legitimate cross-references like "`LcovParseDiagnostic` (in `crap4rs`)")
  - inline `//` comments
  - module-level `//!` doc comments
  - the `crates/crap-core/tests/` tree (snapshots reference renamed-but-historical names)
- [ ] Lefthook `ast-purity` pre-push step mirrors the new check (existing pre-push runs the 3 import gates today).
- [ ] A deliberately-violating literal added then removed locally proves the lint fires.
- [ ] Banned-token list documented inline in the workflow step's comment block.

## Context

- Epic: #173
- Surfaced during: #178 / #179 port-surface cleanup PR (deferred to keep PR scope tight)
- Related: #161 (the prior ast-purity hardening that closed the same anti-pattern class via import-level checks)
- Mechanical-enforcement memory: `feedback_documentation-rots-ci-doesnt`

## Discovery

- **Scope of literal banlist**: should it include `oxc`, `syn`, `napi`? Those are adapter-internal mechanics, but they don't surface in `crap-core/src/` today; widening the list before there's a real leak is YAGNI. Recommend keep the list narrow to user-visible coverage/source formats (`LCOV`, `Istanbul`, `TypeScript`) and revisit per actual leak.
- **Regex strategy**: rustdoc comments are the main false-positive vector. The check needs a multiline pattern that:
  - Matches inside `format!(...)`, `concat!(...)`, `#[error("...")]` attribute strings, and bare `"..."` literals
  - Skips lines starting with `//` / `///` / `//!` (token-level skip via `^[[:space:]]*//`)
  - Alternative: scope to inside `"..."` string literals via a more structural parser, but that's a bigger lift than a grep
- **PCRE2 vs POSIX**: ripgrep is already a hard dep of the CI job (used by grouped-import check). PCRE2 lookbehinds can encode "not preceded by `//`" cleanly; POSIX awk would need a two-pass strip-comments-then-grep.
- **CHANGELOG copy**: the lint name (`ast-purity-strings`?) and CI step name should match the convention of the three existing steps.

cmbays:
## Summary

- Adds a 4th step to the existing `ast-purity` CI job that bans the proper-noun forms `LCOV`, `Istanbul`, `TypeScript`, and `LCOV-flavou?red` from string literals in `crates/crap-core/src/`. Case-sensitive substring-within-string-literal scan + the same awk comment-filter the #161 adapter-name literal check uses, so rustdoc and inline comments stay allowed.
- Mirrors to `lefthook.yml`'s pre-push so violations fail at `git push` time, not at CI time (supply-chain-hygiene convention: "documentation rots; CI doesn't").
- Prerequisite cleanup: drops the literal token `"LCOV"` from the two generic `eprintln!` messages in `cli/mod.rs` (`warn_if_issues` / `print_diagnostics`) so the baseline scan lands clean. These two messages were emitted by a `<P: ParseDiagnostic>`-parameterized helper, so the hard-coded `"LCOV"` was already wrong by design under `crap4ts`/Istanbul — exactly the regression class this layer catches.

Closes #261

## Context

`crap-core` is the language-agnostic domain of the workspace. Existing `ast-purity` layers ban AST-library imports (L1/L2) and direct deps (L3) plus adapter-name literals like `"crap4rs"`/`"crap4ts"` (#161). Layer 4 closes the source-format-vocabulary gap: a change can leak adapter knowledge in user-facing prose without importing or naming anything otherwise banned. Inline comment in `ci.yml` documents the design (including the `rg -E` flag pitfall — ripgrep's `-E` is `--encoding`, NOT grep's `--extended-regex`).

Pipeline: `crap-rs/20260524-ast-purity-strings`. Shape phase locked four decisions (narrow banlist; reuse #161 awk filter; case-sensitive proper-noun match; step name mirrors #161). Build phase deviation: shape's "zero hits" empirical baseline was a false negative (the malformed `rg -E` flag), surfaced two real pre-existing leaks at `cli/mod.rs`, and motivated the case-sensitive choice (lowercase `"lcov.info"` filename literals ≠ user-vocabulary leak). User-approved deviation; both leaks fixed in this PR.

## Test plan

- [x] `crates/crap-core/src/` scans clean under the new lint (baseline + comment-only references preserved)
- [x] Spike test — temporary `const _: &str = "LCOV parse failed";` injection causes the lefthook step to exit 1 with the violation printed
- [x] Existing L1-L3 ast-purity steps unchanged and still green
- [x] `cargo nextest run --workspace --all-targets`: 1111/1111 pass (including the 11 `verbose_integration` tests whose assertions on the renamed strings were updated)
- [x] `cargo insta test --workspace --check`: clean (no snapshot drift)
- [x] `cargo clippy --workspace --all-targets -- -D warnings`: no issues
- [x] CI full battery (verified by the lefthook pre-push battery on `git push`)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -798,17 +798,31 @@ jobs:
         # NOT `--extended-regex` (which is grep's `-E`). Bare
         # invocation `rg PATTERN PATH` is the right shape.
         #
+        # Tool-presence check: `rg` must be on PATH or the gate
+        # would silently pass via `|| true` (the false-sense-of-
+        # security trap). On `ubuntu-latest` ripgrep is preinstalled.
+        #
         # Doc/inline/block comments are allowed to mention these
         # tokens (the awk filter skips lines whose content starts
-        # with //, /*, *). Same comment-filter shape as the #161
-        # adapter-name literal check above; only the token list and
-        # the `"[^"]*X[^"]*"` candidate scan
-        # (substring-within-string-literal) differ.
+        # with `//`, `/*`, `* ` followed by space, or bare `*` at
+        # end of line — covering `///`, `//!`, `/* … */` openers,
+        # and ` * ` / ` *` continuation lines without swallowing
+        # Rust deref-assignment like `*ptr = "LCOV"`).
+        #
+        # The bare `*` form (no space) is intentionally NOT
+        # comment-skipped — Rust comment continuation uses ` * `
+        # (space-asterisk-space) or a bare ` *` (space-asterisk-EOL).
+        # Deref-assign idioms (`*ptr = "Istanbul";`) must still
+        # trip the lint.
         #
         # No untrusted input interpolated; `rg`/`awk` patterns are
         # literal static strings.
         run: |
           set -euo pipefail
+          if ! command -v rg >/dev/null 2>&1; then
+            echo "::error::ripgrep (rg) is required by the ast-purity-strings lint but is not on PATH"
+            exit 1
+          fi
           hits=$(rg -n '"[^"]*(LCOV|Istanbul|TypeScript|LCOV-flavou?red)[^"]*"' crates/crap-core/src 2>/dev/null || true)
           if [ -z "$hits" ]; then
             echo "crap-core adapter-vocabulary literals: clean"
@@ -819,7 +833,7 @@ jobs:
               content = ""
               for (i = 3; i <= NF; i++) content = content (i > 3 ? ":" : "") $i
               gsub(/^[ \t]+/, "", content)
-              if (content !~ /^(\/\/|\/\*|\*)/) print
+              if (content !~ /^(\/\/|\/\*|\* |\*$)/) print
             }
           ')
           if [ -n "$non_comment_hits" ]; then
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -674,6 +674,19 @@ jobs:
       - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
         with:
           persist-credentials: false
+      - name: Install ripgrep
+        # `ubuntu-latest` runner images do not ship ripgrep on every
+        # generation (empirically observed via #261's `command -v rg`
+        # guard tripping on this PR's run). Three steps in this job
+        # shell `rg` — the L2 grouped-import check, the L3.5 adapter-
+        # name literal check (#161), and the L4 adapter-vocabulary
+        # check (#261). The L2 and L3.5 steps use the
+        # `if rg ...; then`/`|| true` patterns that silently treat a
+        # missing `rg` as "no matches found" and exit 0; the L4 step
+        # added in #261 surfaces this loudly via `command -v rg`.
+        # Explicit install closes the gap so all three layers
+        # actually execute. ~10-15s wall-clock overhead per CI run.
+        run: sudo apt-get update && sudo apt-get install -y ripgrep
       - name: Reject AST-library imports in crap-core (leading-token form)
         # Scope: crap-core ONLY. oxc/swc imports are legitimate in
         # crates/crap4ts/ once that lands; crap4rs uses syn. Do NOT
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -775,6 +775,59 @@ jobs:
             exit 1
           fi
           echo "crap-core adapter-name literals: clean (comment-only references allowed)"
+      - name: Reject adapter-vocabulary string literals in crap-core
+        # Layer 4 of crap-core's purity gate. Layers 1-3 (above) ban
+        # `syn`/`oxc`/`tree_sitter`/etc. at the import + direct-dep
+        # level. This layer bans the vocabulary of adapter-specific
+        # coverage source formats from appearing inside user-facing
+        # string literals (`format!`, `#[error("…")]`, bare `"…"`).
+        #
+        # Banlist (case-sensitive, proper-noun forms): `LCOV`,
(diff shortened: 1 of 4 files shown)


cmbays:
## Review-finding disposition

Thanks to @gemini-code-assist for two medium-severity findings. Both addressed in `be18e7c`.

### Gemini #1 (medium) — `rg ... 2>/dev/null || true` silently passes if `rg` is missing
**Paraphrased**: the combination of stderr-suppression (`2>/dev/null`) and exit-status-swallowing (`|| true`) means a missing or broken `rg` produces an empty `hits` variable, the gate exits 0, and the purity check is silently disabled — a false sense of security.

**Disposition**: **Fixed** in `be18e7c` (commit "fix(ci): #261 — address gemini review findings on ast-purity-strings"). Added up-front `command -v rg` check that fails loudly (`::error::ripgrep (rg) is required…` + `exit 1`) before the suppressed pipeline can mask the failure. Both `ci.yml` and `lefthook.yml` mirrors carry the check.

Ironic echo of the same false-negative-baseline trap that bit the shape phase: ripgrep's `-E` is `--encoding` (not grep's `--extended-regex`), and `2>/dev/null` swallowed the flag-mismatch error. New memory `feedback_rg_E_is_encoding_not_extended_regex` captures the broader pattern.

### Gemini #2 (medium) — awk filter `^(\/\/|\/\*|\*)` too aggressive on bare `*`
**Paraphrased**: Rust deref-assignment idiom `*ptr = "LCOV";` after leading-whitespace strip becomes `*p = "LCOV…";`, which matches `^\*` and gets skipped — the banned literal bypasses the lint. Filter is too broad relative to its "skip block-comment continuation" intent.

**Disposition**: **Fixed** in `be18e7c`. Tightened the alternation to `^(\/\/|\/\*|\* |\*$)`: `*` is only treated as comment-continuation when followed by a **space** (`* foo` doxygen-style continuation) or at **end of line** (bare ` *` on its own line in a `/* … */` block). Verified empirically with the spike Gemini's example calls out:

```rust
*p = String::from("LCOV parse failed");
```

…now correctly fires with exit 1 and the violation printed. Under the old filter it would have been skipped.

Residual edge case acknowledged: multi-line Rust expressions where a continuation line legitimately starts with `* foo` (e.g., `let x = a\n    * b`) are still skipped. Gemini's review acknowledged this trade-off; per-statement multi-line `*` lead-in without explicit grouping is rare in crap-core's pure-domain code, and closing it would require a Rust parser rather than a regex. Accepted.

### Sibling-layer follow-up

The #161 adapter-name literal check above carries the same two structural patterns (`|| true` mask + bare-`*` aggressive skip). Filed [#287](https://github.com/breezy-bays-labs/crap-rs/issues/287) (`type:chore` + `priority:soon`) to apply the same hardenings there so L3.5 and L4 stay aligned. Doing both in this PR would have widened the scope past the original #261 charter; the fast-follow keeps the convention consistent.

### CodeRabbit

Rate-limited on PR open (1 review/hour); did not deliver findings before this disposition. Re-trigger via `@coderabbitai review` will be honored if needed before merge.

cmbays:
## CI follow-up: ripgrep was missing from `ubuntu-latest`

Pushing the gemini-review-fix commit (`be18e7c`) caused the **ast-purity job to fail in CI** — the new `command -v rg` guard tripped because `ubuntu-latest`'s current runner image does **not** ship ripgrep.

This was the same false-negative-baseline trap that motivated the gemini #1 finding in the first place. Two pre-existing rg-using layers in the same job were silent-passing:

- L2 (#145) — `if rg -nU --pcre2 ...; then` — `rg: command not found` exits non-zero, `then` branch skipped, step prints "clean" and exits 0.
- L3.5 (#161) — `hits=$(rg ... 2>/dev/null || true)` — masks "command not found" identically.

Both had been no-ops in CI on this runner image since they shipped; the L4 `command -v rg` guard was the first to surface it loudly.

**Fix landed in `3c80709`** — explicit `sudo apt-get install -y ripgrep` at the top of the ast-purity job, before any rg-using step. ~10-15s wall-clock overhead per run; small price for actually-functional gates.

[#287](https://github.com/breezy-bays-labs/crap-rs/issues/287) updated to track (a) propagating the tightened awk filter to L3.5 and (b) auditing the rest of the workflow for the same silent-pass shape elsewhere. The explicit `apt-get install` is the **empirical fix**; the convention alignment in #287 makes the L2/L3.5/L4 shapes uniform after the install is in place.

This finding ranks with the value of the gemini review itself — without that finding I would have shipped L4 with the same silent-pass shape, and the L2/L3.5 silent-pass would have stayed hidden indefinitely. Thanks again @gemini-code-assist.

Issue #260, opened 2026-05-23
feat(crap-core): askama-templated HTML + markdown reporters + revamped HTML report design


cmbays:
## Summary

Migrate the HTML and Markdown reporters from string-concatenation to askama compile-time templates, and ship a revamped HTML report design. Claude Design code for the new HTML report exists locally (reduces design effort).

## Acceptance Criteria

- [ ] askama added as a crap-core dependency; templates live under `crates/crap-core/templates/`
- [ ] HTML reporter migrated to askama (HTML snapshot drift documented + accepted in PR body)
- [ ] Markdown reporter migrated to askama (snapshot drift documented if any)
- [ ] New HTML report design landed: templates + CSS, replacing the existing layout (using the existing Claude Design source)
- [ ] crap4rs + crap4ts both pick up the new reporters via the shared crap-core entry points
- [ ] Wire-envelope canaries (`wire_envelope_crap4rs`, `wire_envelope_crap4ts`, `wire_envelope_crap4ts_branches`) stay byte-identical (envelope is JSON; reporter changes shouldn't touch it — if they do, hard investigation trigger)
- [ ] Existing reporter behaviors preserved: `--format html` and `--format markdown` produce valid output for the same inputs as today
- [ ] Full gate battery green: `cargo +1.93 check --workspace --all-targets` + `cargo fmt --check` + `cargo clippy --workspace --all-targets -- -D warnings` + `cargo doc -D warnings` + `cargo deny check` + `cargo nextest run --workspace --all-targets`
- [ ] crap-core minor bump if reporter public API drifts; patch bump if internal-only

## Context

- Epic: #173 (pre-GA backlog clearance — part of the cluster cleared before W4.1b #193 GA)
- Pipeline: `crap-rs/20260513-typescript-adapter`
- Touches: `crates/crap-core/src/adapters/reporter/html.rs` + `crates/crap-core/src/adapters/reporter/markdown.rs` (and any shared reporter scaffolding)
- Lower-effort than usual: Claude Design output for the new HTML report exists locally and only needs translation into askama templates + CSS

## Discovery

- Pin askama version via Context7 / `cargo search askama` (askama 0.12 vs 0.13 — verify current stable)
- Whether the markdown reporter benefits from askama (it's simpler than HTML) or should stay hand-rolled to keep template surface tight
- HTML snapshot redact strategy: which fields are volatile (timestamps, `tool_version`) and which are content
[...]

PR #311, opened 2026-05-24
feat(crap-core): #306 — HTML reporter delta tab (Current vs baseline view)

cmbays:
## Summary

- Adds the optional **delta tab** to the Sakura HTML reporter that #260 landed. When the operator passes `--baseline <envelope>`, the report renders a tabs nav between the topbar and the body, the **Current** panel opens by default, and a second `<div class="tab-panel" data-tab="delta">` follows with a 4-tile delta KPI grid plus per-category Regressions / Improvements / New-functions tables.
- `format_html` widens to accept `Option<&DeltaView<'_>>` (signature drift; minor bump `crap-core 0.5.0 → 0.6.0`).
- The no-baseline render is **byte-identical** to v0.5.0 — proven via md5 match on the unchanged snapshot file, not just snapshot-test pass.

Closes #306

## What ships

| File | Change | Rationale |
|---|---|---|
| `Cargo.toml` | bump | `crap-core` workspace dep `0.5.0 → 0.6.0` |
| `crates/crap-core/Cargo.toml` | bump | Package version `0.5.0 → 0.6.0` (signature drift) |
| `crates/crap-core/CHANGELOG.md` | +33 | `## [0.6.0]` entry |
| `crates/crap-core/src/adapters/reporters/html.rs` | extended | `format_html` widens; new `DeltaPanel`, `DeltaKpi`, `DeltaRow` structs + `build_delta_panel` projection; 13 new TDD tests; `Box<DeltaPanel>` on `HtmlReport` keeps `Option::None` arm cheap |
| `crates/crap-core/src/cli/mod.rs` | +3 LOC | `FormatArg::Html` dispatch threads `delta_view` |
| `crates/crap-core/templates/html_report.html` | extended | Tabs nav + delta panel + delta CSS layer + tab-switcher IIFE — all gated on `{% if has_delta %}` |
| HTML snapshot (new) | added | `full_html_with_delta_snapshot.snap` |
| HTML snapshot (no-baseline) | **unchanged on disk** | md5 `d51a01255c7919f60f9b7928d3ced7ee` at HEAD~1 and HEAD |

## No-baseline byte-identical proof

The orchestrator's prompt warned about the #260 trap: `insta` trims trailing whitespace on compare, so a passing snapshot test does NOT prove byte equality. The strictest proof is a direct hash of the snapshot files (`insta` only rewrites a snap on mismatch, so an unmodified file hash IS byte-identity):

```
[...]

PR #312, opened 2026-05-25
feat(ci): #294 — HTML report wiring (html-report input + per-language artifact + sticky link)

### LINKED ISSUE #294: feat(ci): HTML report wiring (html-report: true input + artifact upload + sticky-comment link)
## Summary

New optional `html-report: true` input on the composite scorecard action. When `true`: the action runs `<bin> --format html` (the format from #260 Askama HTML reporter — must land first) after the existing JSON/markdown/scorecard-row invocations; uploads the resulting HTML as a workflow artifact via `actions/upload-artifact`; injects a download link into the sticky comment body ("📊 [Download full HTML report](artifact-url)"). Per-language in multi-language mode (from PR β): two HTML artifacts (`crap4rs-report.html` + `crap4ts-report.html`) and two links in the sticky comment.

Lands as PR γ of the ergonomics epic. **HARD-depends on #260** — the `--format html` output format must exist before the action can pass `--format html` to either adapter.

## Acceptance Criteria

- [ ] `html-report: true` triggers HTML rendering + artifact upload + sticky-comment link injection
- [ ] `html-report: false` (default) preserves current action behavior exactly (no new artifact, no sticky link change — additive, zero behavioral change for existing consumers)
- [ ] Artifact filename per language: `crap4rs-report.html` (single Rust dispatch) or `crap4ts-report.html` (single TS dispatch) or both in multi-language mode
- [ ] Artifact uploaded via `actions/upload-artifact` SHA-pinned per AGENTS.md "Supply-chain hygiene" rules 7-10
- [ ] Sticky comment includes "📊 [Download full HTML report](artifact-url)" link below the scorecard markdown
- [ ] Per-language in multi-language mode: two separate artifacts (one per language) + two distinct links in the sticky comment (e.g. "📊 [Download full HTML report — Rust](rust-url) · [TypeScript](ts-url)")
- [ ] Default `actions/upload-artifact` retention (90 days) — documented in action README; no override input this PR (keep input surface lean; if real consumer need surfaces post-ship, file additive `retention-days` input as follow-up)
- [ ] CI smoke verifies: artifact upload succeeds; sticky-comment link renders; downloading the artifact + opening the HTML in a browser shows the rendered scorecard (smoke can `unzip` + `grep` for known scorecard markers — full Playwright-style render-verification is overkill for the smoke)
- [ ] CI green on linux-x86 + macos-arm + macos-x86

## Context

- **Epic**: #291
- **Blocked-by**:
  - PR α (preset surface — `html-report` is a new preset-tier input; lives alongside `threshold-preset`/`run-mode`/`gate-mode`/`languages`)
  - **#260** (Askama HTML reporter — `--format html` doesn't exist until #260 ships)
- **Delivery model documented**: workflow artifact + sticky-comment link. NOT GH Pages (too heavy for per-PR review use case). NOT external hosting (requires credentials; not consumer-neutral). PR comments don't render arbitrary HTML/CSS/JS (markdown-only), so inline-HTML-in-comment is structurally impossible. Artifact + link is the realistic shape.
- **Shaping doc**: `~/Github/ops/workspace/crap-rs/20260513-typescript-adapter/shaping-ergonomics.md`

## Discovery

What needs resolving during /plan or in-PR:

- **Artifact naming convention** when multi-language — current spec: `crap4rs-report.html` + `crap4ts-report.html`. Confirm vs alternative (`crap-report-rust.html` + `crap-report-typescript.html` for parallelism with the `row-json-rust` output naming).
- **Sticky-comment link injection shape** — exact markdown: single-link line ("📊 [Download full HTML report](url)") or per-language multi-link ("📊 [Download full HTML report — Rust](rust) · [TypeScript](ts)"). Lock during PR.
- **Artifact URL availability in the same workflow run** — `actions/upload-artifact@v4` returns the artifact-url as a step output; verify timing (artifact uploaded BEFORE sticky-comment step runs).
- **Per-language artifact in single-language mode** — `languages: rust` + `html-report: true` produces one artifact + one link, not two. Lock conditional logic.
- **Retention override input** — DEFERRED per acceptance criterion; document as "may be added as additive follow-up if real consumer need surfaces."
- **May spawn sub-issues**: post-ship `retention-days` input; post-ship HTML-report variant for `comment-mode: none` consumers (where there's no sticky to inject the link into — output the URL as an additional action output instead).

cmbays:
## Summary

- Adds `html-report: true` input on the public composite scorecard action. When `true`, the action runs `<bin> --format html` (per the Askama reporter that landed in `crap-core 0.5.0` + delta-tab activation in `0.6.0`), uploads the rendered HTML as a workflow artifact via `actions/upload-artifact`, and — when `comment-mode: sticky` — appends a download link to the sticky-comment body.
- Per-language in multi-language mode (from PR β): two artifacts (`crap4rs-report-<suffix>` + `crap4ts-report-<suffix>`) and two distinct links in the sticky comment.
- Adds 2 new action outputs (`html-artifact-url-rust` + `html-artifact-url-typescript`) so aggregator workflows (`comment-mode: 'none'`) can read the URLs programmatically without scraping the sticky comment.
- Adds `html-artifact-name-suffix` input (default `'-${{ runner.os }}'`) to prevent matrix-strategy collision on upload-artifact@v4+'s "same artifact name fails fatally on second leg" trap.
- Default `false` — zero behavior change for every existing consumer; back-compat is byte-identical (verified locally — see "Wire-snapshot canary discipline" below).

Closes #294

## What ships

| File | Change | Rationale |
|---|---|---|
| `.github/actions/scorecard/action.yml` | +201 LOC | 2 new inputs (`html-report` + `html-artifact-name-suffix`); 2 new outputs (`html-artifact-url-{rust,typescript}`); HTML render arm inside the per-language loop; 2 new upload steps (per language, `if:` gated on resolved-language set); new `Compose sticky message` step that appends the link(s) to the markdown body when `html-report: true`; `Post sticky PR comment` rewired to read `steps.compose.outputs.message` |
| `.github/actions/scorecard/README.md` | +109 LOC | New `## HTML report` section between `## Multi-language` and `## Minimal usage` (inputs, sticky-vs-aggregator pattern, artifact naming + 90-day retention rationale, per-language behavior). Inputs + Outputs tables expanded with the new entries |
| `.github/workflows/ci.yml` | +86 LOC | New `HTML report spot-check` step in `scorecard-smoke-rust` exercising the full path: `html-report: true` + sticky on PR events, then asserts `html-artifact-url-rust` output non-empty + artifact uploaded (`gh api`) + HTML file parses (`</html>` + `<h1>crap4rs report</h1>` markers). Skips the existing `scorecard-smoke` internal composite intentionally — that composite predates `html-report` and doesn't surface the new outputs |

## Implementation deviations from the impl plan

Surfaced explicitly per the orchestrator's "Plan deviations" protocol. ε caught three; γ caught two.

**Plan deviation #1 — `Compose sticky message` step is structurally new (not a modification of the existing sticky step).** The impl plan implied modifying `Post sticky PR comment`'s `message:` value in place. Actual structure: `Post sticky PR comment` consumes a sticky-comment uses-action whose `with: message: ${{ … }}` resolves at step-init time, so the only way to inject the artifact URLs is to compose the body in a separate step that runs AFTER the upload steps (which write `artifact-url`) and BEFORE the sticky step. Resolution: new `Compose sticky message` step that reads both upload-step `artifact-url` outputs + the markdown body, emits a `message` output via heredoc, and the sticky step rewires its `message:` to `${{ steps.compose.outputs.message }}`. Advisor catch during pre-build review.

**Plan deviation #2 — `html-artifact-name-suffix` empty-default uses GHA ternary (not bash-side fallback).** The impl plan's verbatim used `${{ inputs.html-artifact-name-suffix || format('-{0}', runner.os) }}` which works on truthy-falsy semantics. Actual GHA evaluates `''` as falsy in that ternary (verified empirically by the false-default semantics PR α established), but the explicit equality form `inputs.html-artifact-name-suffix == '' && format('-{0}', runner.os) || inputs.html-artifact-name-suffix` is more defensive — surfaces clearly that empty-string is the trigger for the OS-suffix default. Doc-string for the input explains the substitution.

## Acceptance gates verified

| Gate | Status |
|---|---|
| `cargo build --workspace` | ✓ pass |
| `cargo nextest run --workspace --all-targets` | ✓ **1156/1156 pass** |
| `cargo fmt --check` | ✓ clean |
| `cargo clippy --workspace --all-targets -- -D warnings` | ✓ clean ("No issues found") |
| `cargo +1.93 check --workspace --all-targets` (MSRV) | ✓ clean |
| `cargo deny check` | ✓ advisories / bans / licenses / sources ok |
| `cargo doc --workspace --no-deps` (with `RUSTDOCFLAGS=-D warnings`) | ✓ clean |
| AST-purity grep (`use syn`/`use oxc`/`use lcov` in `crap-core/src/`) | ✓ clean (no source touched) |
| `scripts/bdd-tracked-lint.py` | ✓ ok (246 deferred / 384 scanned) |
| `scripts/mutants-skip-lint.py` | ✓ ok (3 `--skip` tokens cover all adapter-bin shelling tests) |
| Wire-envelope canaries (×3) byte-identical | ✓ no drift (action.yml + README + ci.yml don't reach `crates/crap-core/`) |
| `actionlint .github/workflows/ci.yml` | ✓ clean |
| `shellcheck` on the new `Compose sticky message` bash | ✓ clean |
| `pipx run 'zizmor>=1.5,<2' .github/` | ✓ "No findings to report. Good job!" (1 ignored, 20 suppressed) |
| `python3 -c "import yaml; yaml.safe_load(open('.github/actions/scorecard/action.yml'))"` | ✓ valid YAML |
| Heredoc emission dry-run (mimics #260's trailing-newline fix) | ✓ EOF terminator on its own line in all 3 scenarios (no-html, single-lang+url, multi-lang+2-urls) |
| CI matrix (linux-x86 / macos-arm / macos-x86) | pending CI run |

## Wire-snapshot canary discipline

Byte-identical (no changes under `crates/crap-core/`). PR declares **no drift** — `git diff --stat crates/crap-core/tests/snapshots/` returns empty.

## actions/upload-artifact SHA pin + version

Pinned to `043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1` (the in-repo SHA matching `.github/workflows/ci.yml:194`, `.github/workflows/release-plz.yml:243`, and `.github/workflows/release-plz.yml:402`). Three pins now reference the same SHA — consistency over chasing dependabot's latest; a future bump rebases all three in one weekly group per `.github/dependabot.yml`'s grouped-updates convention. AGENTS.md supply-chain hygiene rules 7-10 verified (SHA pin + `# vX` trailer; no `persist-credentials` issues since upload-artifact doesn't checkout; not a tag-triggered workflow so cache-poisoning rules don't apply).

## artifact-url synchronous availability (Context7 verification)

The Context7 lookup against `actions/upload-artifact@v7.0.1/action.yml` confirmed that:
- `artifact-url` is an exposed step output ("A download URL for the artifact")
- It is "populated when the artifact upload succeeds" and "empty if the artifact upload failed"
- It is "available synchronously after the step completes within the same job"

So the `Compose sticky message` step that runs immediately after the upload steps can safely read `steps.upload-rust.outputs.artifact-url` + `steps.upload-typescript.outputs.artifact-url` and get the URLs without polling. The same property carries through to the action-level `html-artifact-url-{rust,typescript}` outputs, which downstream `if:` / `env:` consumers can read in the next step of a calling workflow.

The doc-string on both outputs notes that the URL "only resolves for requests authenticated with GitHub" — the smoke assertion therefore uses `gh api repos/$repo/actions/runs/$run/artifacts` to verify the upload (one server-side API call against the authenticated `GH_TOKEN`), rather than attempting to `curl` the artifact URL directly inside the same job.

## 90-day retention default + rationale for no override input

`actions/upload-artifact@v4+`'s `retention-days` defaults to the workflow/repository default of **90 days** when unspecified. This PR does NOT expose a `retention-days` input on the composite action; the README documents the 90-day default and reserves the override as a follow-up additive input if real consumer friction surfaces. Rationale (per impl plan L597 + shape-doc intra_shape_lock #5):

- **90 days covers the "review the PR within a quarter" use case.** PR-review HTML reports are ephemeral; nobody re-opens a closed PR three months later expecting the artifact to still be downloadable. The default already exceeds typical PR-review windows.
- **Keep the input surface lean.** Each new input is a permanent maintenance surface (test matrix entry, README row, validation logic in `Resolve presets`). Skipping inputs without demonstrated demand keeps the action approachable.
- **Additive follow-up is cheap.** If a consumer surfaces real friction (e.g. compliance retention requirement, periodic-job aggregation that re-reads artifacts), an additive `retention-days` input lands as a one-line passthrough to `actions/upload-artifact`'s `retention-days:` input.

## comment-mode: none aggregator pattern

When a consumer sets `comment-mode: 'none'` + `html-report: true`, the action:
- **Still uploads** the HTML artifact(s) per language (the upload steps are gated on `inputs.html-report == 'true' && (steps.lang.outputs.language == 'rust' || …)`, NOT on `comment-mode`)
- **Still surfaces** the artifact URL(s) via `html-artifact-url-{rust,typescript}` outputs (consumer reads programmatically)
- **Skips the sticky link injection** (the sticky step is gated on `inputs.comment-mode == 'sticky'`)
- The `Compose sticky message` step still runs (no `comment-mode` gate) and produces a `message` output — but no consumer reads it when `comment-mode: 'none'`, so the side-effect is null

This matches the action's stated primary use case (aggregator workflows composing multiple metric rows into one comment).

## Local verify command list

```bash
cargo fmt --check
cargo clippy --workspace --all-targets -- -D warnings
cargo +1.93 check --workspace --all-targets
cargo nextest run --workspace --all-targets
RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps
cargo deny check
python3 scripts/bdd-tracked-lint.py
python3 scripts/mutants-skip-lint.py
actionlint .github/workflows/ci.yml
pipx run 'zizmor>=1.5,<2' .github/
```

## Test plan

- [ ] CI matrix green on linux-x86 / macos-arm / macos-x86 for `test`, `clippy`, `fmt`, `msrv`, `deny`, `docs`, `zizmor`, `bdd-tracked-lint`, `mutants-skip-lint`
- [ ] `scorecard-smoke-rust` green — the new `HTML report spot-check` step asserts: (a) `html-artifact-url-rust` non-empty, (b) `html-artifact-url-typescript` empty, (c) artifact uploaded under `crap4rs-report-html-spotcheck`, (d) HTML file at `$RUNNER_TEMP/crap4rs-report.html` contains `</html>` + `<h1>crap4rs report</h1>`
- [ ] `scorecard-smoke-ts` green (back-compat — `html-report` defaults to false; this job's behavior unchanged)
- [ ] `scorecard-smoke-multi-lang` green (back-compat — same; the 8 existing spot-checks don't pass `html-report`, so they exercise the false-path which is byte-identical to today)
- [ ] Per-PR `mutants` job (`view.rs` leg) green
- [ ] gemini-code-assist + CodeRabbit reviews; orchestrator will address any findings per repo convention (fix push + finding→fix mapping comment)
- [ ] On merge: verify #294 auto-closed via the `Closes #294` keyword (plain prose, not in backticks, per `feedback_closes-keyword-not-in-backticks`)

### DIFF
--- a/.github/actions/scorecard/README.md
+++ b/.github/actions/scorecard/README.md
@@ -183,6 +183,111 @@ containing the combined `markdown` output (both per-language sections
 stacked). Aggregator workflows reading the split outputs and
 re-rendering elsewhere should still set `comment-mode: 'none'`.
 
+## HTML report
+
+For PR reviews that benefit from the full file-by-file scorecard (KPI
+tiles, distribution bar, file cards with sortable tables, optional dark
+mode) — set `html-report: true`. The action renders an HTML report via
+`<bin> --format html`, uploads it as a workflow artifact, and (when
+`comment-mode: sticky`) appends a download link to the sticky comment
+body.
+
+When `baseline:` is supplied alongside `html-report: true`, the
+rendered HTML automatically activates the Delta tab (Current vs
+baseline) — no separate input needed.
+
+```yaml
+- uses: breezy-bays-labs/crap-rs/.github/actions/scorecard@<sha>
+  with:
+    coverage: lcov.info
+    baseline: /tmp/baseline.json
+    html-report: true
+    comment-mode: 'sticky'
+```
+
+| Input | Default | Notes |
+|---|---|---|
+| `html-report` | `false` | When `true`, render + upload + (sticky-only) inject link |
+| `html-artifact-name-suffix` | `''` → `-${{ runner.os }}` | Disambiguates matrix-strategy uploads |
+
+### Why artifact + link, not inline HTML
+
+GitHub PR comments are rendered as **markdown only** — `<script>`,
+`<style>`, inline `data:` images, and most arbitrary HTML/CSS are
+stripped or sanitized. Embedding the rendered HTML in the comment
+body therefore isn't structurally available. The artifact + link
+pattern works with GitHub's existing UI: reviewers click the link,
+GitHub authenticates them, the artifact downloads as a zip, and the
+full report opens in any browser (the renderer is single-file +
+self-contained, no external assets).
+
+### Artifact naming + retention
+
+- **Filename per language**: `crap4rs-report-<suffix>` (Rust) and/or
+  `crap4ts-report-<suffix>` (TypeScript). The default suffix is
+  `-${{ runner.os }}` so the common matrix-over-OS case produces
+  cleanly distinguishable names (`crap4rs-report-Linux`,
+  `crap4rs-report-macOS`). Override via `html-artifact-name-suffix`
+  for matrix dimensions beyond OS (`-Linux-x86_64`, `-Linux-arm64`,
+  …). `actions/upload-artifact@v4+` fails fatally on the second
+  matrix leg attempting to upload to the same artifact name, so a
+  unique suffix per leg is required for matrix-strategy workflows.
+- **Retention**: 90 days (GitHub's default `actions/upload-artifact`
+  retention window — long enough to cover the typical "review the PR
+  within a quarter" use case). The action does not expose a
+  retention override input in this release; if real consumer friction
+  surfaces we can ship an additive `retention-days` input as a
+  follow-up.
+
+### Per-language behavior in multi-language mode
+
+When `languages: rust,typescript` (or `all`) is set alongside
+`html-report: true`, the action uploads **two artifacts** (one per
+language) and the sticky comment carries **two links**:
+
+```
+📊 [Download full HTML report — Rust](<rust-url>) · [TypeScript](<ts-url>)
+```
+
+The two artifact URLs are also surfaced as action outputs
+(`html-artifact-url-rust` + `html-artifact-url-typescript`) so an
+aggregator workflow running `comment-mode: 'none'` can read the URLs
+and render its own link block. When a language is not in the
+resolved set, its respective output is empty.
+
+```yaml
+- uses: breezy-bays-labs/crap-rs/.github/actions/scorecard@<sha>
+  id: crap
+  with:
+    languages: rust,typescript
+    coverage:    lcov.info
+    coverage-ts: coverage-final.json
+    src:    crates/
+    src-ts: packages/
+    html-report: true
+    comment-mode: 'none'             # aggregator pattern — no sticky
+
+- name: Render combined report
+  run: |
+    echo "Rust:       ${{ steps.crap.outputs.html-artifact-url-rust }}"
+    echo "TypeScript: ${{ steps.crap.outputs.html-artifact-url-typescript }}"
+```
+
+### Aggregator pattern — `comment-mode: none` + `html-report: true`
+
+The action still uploads both artifacts and surfaces their URLs as
+outputs even when `comment-mode: 'none'` — there is no sticky
+comment to inject a link into, but the URLs are still available
+programmatically. Aggregator workflows that compose multiple metric
+rows into one comment can read the HTML-artifact URL outputs and
+render their own link wherever appropriate (sticky comment, Check
+Run summary, Slack message, …).
+
+The artifact URLs only resolve for requests **authenticated with
+GitHub** (per the GH API contract on artifact download URLs), so the
+links are meant for human-followable surfaces (sticky comments,
+Check Run summaries) — not in-job `curl` retrieval.
+
 ## Minimal usage — analysis only, no delta
 
 ```yaml
@@ -307,6 +412,8 @@ row; the aggregator owns the comment.
 | `version` | `latest` | crap4rs version to install (`latest` or pinned tag) |
 | `annotations` | `false` | When `true`, emit `::warning` workflow commands so findings render inline on the PR Files Changed tab (see [Inline annotations](#inline-annotations)) |
 | `annotation-limit` | `''` | Cap on emitted annotations when `annotations: true`. Empty defers to the adapter's default (10) or `[output] annotation_limit` from `config`. Range 1..=100 |
+| `html-report` | `false` | When `true`, render `<bin> --format html` as a workflow artifact and (sticky-only) append a download link to the comment body. Per-language in multi-language mode. See [HTML report](#html-report) |
+| `html-artifact-name-suffix` | `''` (→ `-${{ runner.os }}`) | Suffix appended to the HTML artifact name(s) to disambiguate matrix-strategy uploads. See [HTML report — Artifact naming + retention](#artifact-naming--retention) |
 | `threshold-preset` | `''` | (preset) `strict` (8) \| `default` (15) \| `lenient` (25). Derives `threshold`; raw `threshold:` wins on conflict + warning. See [Presets](#presets) |
 | `run-mode` | `''` | (preset) `full` \| `delta` \| `both` — drives baseline expectations. `delta`/`both` require `baseline:` set. See [Presets](#presets) |
 | `gate-mode` | `''` | (preset) `report-only` \| `gate-on-analysis` \| `gate-on-delta` \| `gate-on-both`. Atomic `analysis-gate` + `delta-gate` pair; raw inputs win on conflict + warning. See [Presets](#presets) |
@@ -326,6 +433,8 @@ row; the aggregator owns the comment.
 | `new-violations` | Count of functions exceeding threshold but not in baseline. Multi-language: SUM across languages |
 | `regressions` | Count of modified functions whose CRAP increased above rendering precision. Multi-language: SUM across languages |
 | `threshold-resolved` | The threshold value the action's `Resolve presets` step resolved (preset-derived or raw passthrough). See [Presets — Threshold-sync invariant](#threshold-sync-invariant) |
+| `html-artifact-url-rust` | (html-report) Workflow artifact URL for the rendered Rust HTML report when `html-report: true` and Rust is in the resolved language set. Empty otherwise. See [HTML report](#html-report) |
+| `html-artifact-url-typescript` | (html-report) Workflow artifact URL for the rendered TypeScript HTML report when `html-report: true` and TypeScript is in the resolved language set. Empty otherwise. See [HTML report](#html-report) |
 
 ## Structured row output (`outputs.row-json`)
 
--- a/.github/actions/scorecard/action.yml
+++ b/.github/actions/scorecard/action.yml
@@ -1338,7 +1338,11 @@ runs:
       if: steps.lang.outputs.language == 'rust' || steps.lang.outputs.language == 'typescript' || steps.lang.outputs.language == 'multi'
       shell: bash
       env:
-        MARKDOWN: ${{ steps.run.outputs.markdown }}
+        # Markdown body is read FROM the temp files Run analysis already
+        # wrote (rather than re-piped through an env var) so we don't
+        # tip the 64 KB env-var limit on large projects — gemini catch
+        # on PR #312. Filenames: `${BIN}-scorecard.md` for single-lang,
+        # `scorecard-combined.md` for multi-lang (see L898 + L1050).
         HTML_REPORT: ${{ inputs.html-report }}
         IS_MULTI: ${{ steps.presets.outputs.is_multi }}
         LANGUAGE: ${{ steps.lang.outputs.language }}
@@ -1347,8 +1351,13 @@ runs:
       run: |
         set -euo pipefail
         composed_file="$RUNNER_TEMP/scorecard-sticky-message.md"
-        : > "$composed_file"
-        printf '%s' "$MARKDOWN" > "$composed_file"
+        if [ "$LANGUAGE" = "multi" ]; then
+          cat "$RUNNER_TEMP/scorecard-combined.md" > "$composed_file"
+        elif [ "$LANGUAGE" = "rust" ]; then
+          cat "$RUNNER_TEMP/crap4rs-scorecard.md" > "$composed_file"
+        else
+          cat "$RUNNER_TEMP/crap4ts-scorecard.md" > "$composed_file"
+        fi
 
         if [ "$HTML_REPORT" = "true" ]; then
           # Guard the link append on actual upload success — when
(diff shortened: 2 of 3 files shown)


cmbays:
## Follow-up commit: drop \`gh api\` artifact check from html-report smoke

**Catch**: Pre-PR advisor review surfaced a permissions gap — the original assertion step called \`gh api repos/\$repo/actions/runs/\$id/artifacts\` to verify the upload landed in the workflow's artifact list. That REST endpoint requires \`actions: read\` on the calling job, but \`scorecard-smoke-rust\`'s \`permissions:\` block grants only \`pull-requests: write\` + \`contents: read\`. Once GitHub sees an explicit \`permissions:\` block it becomes a closed allow-list, so the API call would have 403'd on first push.

**Why local gates missed it**: zizmor flags EXCESSIVE permissions (the \`excessive-permissions\` audit), NOT missing ones. actionlint doesn't model REST-API permission requirements either. The failure mode surfaces only on the live runner.

**Resolution**: dropped the \`gh api\` check entirely (option 2 from the advisor's framing). Two reasons it's the better path than option 1 (add \`actions: read\`):

1. **The URL-non-empty check is a stronger signal.** Per \`actions/upload-artifact@v7\`'s contract (Context7-verified during PR pre-flight), \`artifact-url\` is "empty if the artifact upload failed" — so a non-empty URL output IS the upload-success contract. The REST list-artifacts call was duplicating the same signal with weaker semantics ("artifact exists in the run's artifact list").
2. **Keeps the permissions surface minimal.** The job's \`permissions:\` block doesn't expand; the \`actions: read\` scope stays unclaimed for this job.

**This is Plan deviation #3** (gets folded into the PR body's Plan deviations section on next push if reviewers ask).

Commit: \`404cca2\`.

cmbays:
## CI failure → fix mapping (commit `a85ea54`)

The `Scorecard action smoke — TypeScript (analysis-only)` job failed on `404cca2` with:

```
##[error]/home/runner/work/crap-rs/crap-rs/./.github/actions/scorecard/action.yml (Line: 242, Col: 18):
  Unrecognized named-value: 'runner'. Located at position 1 within expression: runner.os
##[error]Failed to load /home/runner/work/crap-rs/crap-rs/./.github/actions/scorecard/action.yml
```

### Root cause

The new `html-artifact-name-suffix` input's `description: |` block-scalar contained the literal string `-${{ runner.os }}` as illustrative syntax explaining what the default expands to. GitHub's composite-action manifest loader **evaluates `${{ ... }}` expressions inside `description:` text** (not just inside `value:` / `with:` / `run:` contexts), so `runner.os` was resolved at action-LOAD time — before any step context exists — and the loader rejected the manifest.

The actual functional use of `runner.os` at `steps.upload-rust.with.name:` (and `steps.upload-typescript.with.name:`) is **correct** — `runner.os` is in scope inside step `with:` blocks. The bug was scoped strictly to documentation prose.

### Fix

Rephrase the description text + a nearby comment to reference `<runner.os>` (literal angle-bracketed placeholder, common docs convention) instead of GHA expression syntax. Same documentation intent, no expression evaluation at load time. Two spots:

| Location | Before | After |
|---|---|---|
| `inputs.html-artifact-name-suffix.description` | `defaults to \`-${{ runner.os }}\`` | `defaults to \`-<runner.os>\` (Linux/macOS/Windows)` |
| Comment above `Upload Rust HTML report` step | `the default \`-${{ runner.os }}\`` | `the default \`-<runner.os>\`` |

### Verification

- `grep -nE '\$\{\{' action.yml` inside inputs/outputs blocks now matches only `value:` fields (the only valid expression context at action manifest top level)
- Behavior unchanged: `steps.upload-{rust,typescript}.with.name:` still resolves `runner.os` correctly at step-runtime

### Lesson PARKed

Add to crap-rs Composite-action gotchas log: **`description:` block-scalars of `inputs:` are evaluated as GHA expressions, not treated as opaque docs**. Anyone documenting GHA syntax inside an input description must escape `${{` (or use angle-bracket placeholder convention like the fix). README.md is unaffected (rendered by GitHub.com as markdown, not parsed by the action manifest loader).

— orchestrator session (parent Claude, model Opus 4.7)

cmbays:
## Gemini review finding → fix mapping (commit `6a3e50a`)

Both gemini findings on commit `a85ea54` addressed.

### Finding 1 — `MARKDOWN` env var risks 64 KB ceiling (HIGH severity, `action.yml:1351`)

**Root cause:** the `Compose sticky message` step passed the full markdown body through an env var (`MARKDOWN: ${{ steps.run.outputs.markdown }}`). GHA runners cap env-var values at ~64 KB on some platforms; a sufficiently large project's scorecard (many functions / multi-language combined body) could blow that ceiling and break the step.

**Fix:** drop the env var entirely. Branch on `$LANGUAGE` and `cat` directly from the temp file that `Run analysis` already wrote:
- `multi` → `$RUNNER_TEMP/scorecard-combined.md` (assembled at L1050)
- `rust`  → `$RUNNER_TEMP/crap4rs-scorecard.md` (single-lang Rust, L898)
- else    → `$RUNNER_TEMP/crap4ts-scorecard.md` (single-lang TypeScript)

These files are guaranteed to exist for the matching `$LANGUAGE` (Run analysis writes them deterministically before Compose runs in step order). Also removes the no-longer-needed `: > "$composed_file"` truncation since `cat > "$file"` creates a fresh file.

### Finding 2 — Wrap markdown link URLs in angle brackets (MEDIUM severity, `action.yml:1373`)

**Root cause:** plain `[text]($URL)` markdown link syntax breaks if `$URL` contains parentheses or spaces. GitHub-Flavored Markdown convention is `[text](<$URL>)` for safety.

**Fix:** wrap all 5 link templates (1 single-language Rust, 1 single-language TypeScript, and 3 multi-language combinations) with `<...>`. Behavior unchanged on the happy path (`actions/upload-artifact@v7`'s URL shape doesn't currently include special chars), but defended against future URL-shape regression.

### Verification

- `grep -nE '\$\{\{ *steps\.run\.outputs\.markdown' action.yml` → 0 matches (env var dropped)
- All 5 link template strings now use `(<$HTML_URL_*>)` form
- File branching covers all `$LANGUAGE` values the surrounding `if:` guard admits (`rust`/`typescript`/`multi`)
- Behavior unchanged on the happy path; defends two failure modes that were latent until a large enough project or special-char URL surfaced them

### CodeRabbit status (per ε lesson)

CodeRabbit's SUCCESS status check on this PR is a false positive — the bot hit the per-user rate limit (`Review limit reached. Refill in 22 minutes`) and produced no findings. Same gotcha as on PR #311; the lesson is parked in the orchestrator's observations log pending escalation on a third occurrence.

— orchestrator session (parent Claude, model Opus 4.7)

PR #350, opened 2026-05-29
chore(ci): #287 — ast-purity layer-4 hardening (command -v rg guard + tighter filter + silent-pass audit)

### LINKED ISSUE #287: chore(ci): #145 + #161 ast-purity lints — apply layer-4 hardenings (command -v rg + tighter * filter) AND audit silent-pass risk
## Summary

The #161 adapter-name literal check (`Reject adapter-name string literals in crap-core` step in `.github/workflows/ci.yml`) and the #145 grouped-import check (`Reject AST-library imports in crap-core (grouped-import form)` step) both carry the same two structural patterns that the gemini review on #286 flagged in the new layer-4 step:

1. Failure of `rg` is silently treated as "no matches found":
   - L2 (#145) uses `if rg ...; then` form — if `rg` exits non-zero (e.g. not on PATH), the `then` branch is skipped and the step prints "clean" and exits 0.
   - L3.5 (#161) uses `2>/dev/null || true` — same false-negative on missing `rg`.
   - Empirically confirmed on PR #286's CI run: `ubuntu-latest` on this runner image does NOT ship ripgrep; **the L2 and L3.5 layers had been no-ops in CI on this image since they shipped.** PR #286's L4 step exposed it via a `command -v rg` guard; commit `3c80709` adds an explicit `apt-get install -y ripgrep` at the top of the ast-purity job to fix all three layers.

2. awk comment-skip pattern `^(\/\/|\/\*|\*)` is too aggressive on bare `*` line-start, so Rust deref-assign idioms like `*ptr = "crap4rs"` would bypass the lint. PR #286's L4 step was hardened to `^(\/\/|\/\*|\* |\*$)`.

This issue tracks (a) propagating the awk-filter tightening to the L3.5 step in `ci.yml` to align with L4, and (b) auditing for any equivalent silent-pass risk elsewhere in the workflow that the explicit ripgrep install on the ast-purity job may have papered over rather than fixed.

## Acceptance Criteria

- [ ] `.github/workflows/ci.yml` "Reject adapter-name string literals in crap-core" step (L3.5) tightens its awk filter `^(\/\/|\/\*|\*)` → `^(\/\/|\/\*|\* |\*$)`.
- [ ] Empirical-fire spike for L3.5: inject `*p = String::from("crap4rs");` in `crap-core/src/`, confirm the lint fires (exit 1 with the violation printed) under the tightened filter.
- [ ] Audit: grep the workflow for any other `2>/dev/null || true` or `if <cmd> ...; then`/`|| true` patterns where a missing tool would silently mean "no matches"; add `command -v <tool>` guards to each.
- [ ] Decide whether the L2 step (#145) needs a tightened filter (it doesn't currently use the comment-skip awk pipeline; the rg PCRE2 expression is bounded to `use { ... }` blocks).
- [ ] CI step name and inline comment block reference the same hardening rationale as layer 4 for consistency.

## Context

- Surfaced during gemini-code-assist review of [#286](https://github.com/breezy-bays-labs/crap-rs/pull/286) (the layer-4 string-literal ban PR closing #261).
- Layer 4 hardenings landed in commits `be18e7c` (gemini findings) + `3c80709` (ripgrep install).
- Related: #145 (grouped-import check), #161 (adapter-name literal check), #261 (layer-4 ban).
- The ripgrep install in `3c80709` (PR #286) is the **empirical fix**; this issue is the **convention alignment** so L3.5 (and L2 if applicable) carry the same guard-and-filter shape as L4 even after the install is in place.

## Discovery

- Should the `command -v rg` check be lifted to a shared step name or a composite action to avoid each ast-purity layer repeating the boilerplate? Probably not worth the indirection for three steps; revisit if a 6th rg-based purity layer ever lands.
- Should `lefthook.yml`'s `ast-purity` command (which currently uses `grep`, not `rg`) be migrated to `rg` for consistency with the rest of the ast-purity family? Separate question; not strictly required by this issue.
- Are there other CI workflows in `.github/workflows/` that shell `rg` without a presence guard? Audit-scope question for this issue.

cmbays:
Propagates the layer-4 (#286) ast-purity hardenings to the two older rg-shelling steps in `.github/workflows/ci.yml`, so all three adapter-string/import gates carry one consistent guard-and-filter shape. Convention alignment on top of the empirical ripgrep-install fix that landed in #286 (commit `3c80709`).

## What changed (`.github/workflows/ci.yml`, ast-purity job only)

**L2 — "Reject AST-library imports in crap-core (grouped-import form)" (#145)**
- Added an up-front `command -v rg` guard. The `if rg ...; then` form treats a missing `rg` identically to "no matches found" and exits 0 — a silent pass.
- **AC4 decision:** L2 gets **no filter change** — it has no awk comment-skip pipeline to tighten; it is a bounded PCRE2 match on `use { ... }` blocks. It gets the guard only.

**L3.5 — "Reject adapter-name string literals in crap-core" (#161)**
- Added the same `command -v rg` guard.
- Tightened its awk comment-skip filter `^(\/\/|\/\*|\*)` → `^(\/\/|\/\*|\* |\*$)` — byte-identical to L4. The over-broad bare `^\*` skip let Rust deref-assign idioms like `*p = "crap4rs";` bypass the lint (they start with `*` then a variable). The tightened form still allows legitimate ` * ` / ` *` block-comment continuation lines.

**`Install ripgrep` step comment (AC5)**
- De-staled: it claimed only L4 guarded `command -v rg`. After this PR all three rg-shelling steps guard. The explicit apt install stays the empirical fix; the per-step guards are defence-in-depth that fail loudly if the install regresses or a runner image drops ripgrep again.

## Audit result (AC3 + AC4)

`rg` is the **only** external-and-possibly-absent tool shelled by the workflow. The other tools are POSIX-core / preinstalled on `ubuntu-latest` and do **not** need presence guards:
- L1 uses `grep -rnE` (POSIX grep, always present), L3 uses `jq` (preinstalled).
- `release-plz.yml` `grep -E '^SF:/'` + `jq`, and `quick-start-smoke.yml` `grep -qF` checks are all always-present tools.

After this PR all three rg-shelling steps (L2, L3.5, L4) carry the guard. `lefthook.yml` needs **no change**: its only rg use is the L4-equivalent `ast-purity-strings` hook (already guarded), and it has no L2/L3.5 mirror.

## Silent-pass evidence (AC2 + AC3)

Drills run with the **exact awk literals** from the workflow against a scratch tree (never committed; working tree showed only `ci.yml` dirty before commit).

**AC2 — filter tightening (L3.5):**

```
### Case A: deref-assign violation  *p = String::from("crap4rs");
  OLD filter ^(//|/*|*):       clean (exit 0)        <- SILENT PASS (the bug)
  NEW filter ^(//|/*|* |*$):   FIRES (exit 1)         <- the fix
      /tmp/.../violation.rs:2:    *p = String::from("crap4rs");

### Case B: legitimate comment refs (///, //, " * ")
  NEW filter ^(//|/*|* |*$):   clean (exit 0)         <- comment allowance preserved
```

**AC3 — rg-missing guard (run under a stripped PATH with a real `"crap4rs"` violation on disk):**

```
### Without guard (current L2/L3.5 shape): rg ... 2>/dev/null || true
  -> reported: clean (SILENT PASS despite real violation)   exit=0

### With guard (the #287 fix): up-front command -v rg
  ::error:: ripgrep (rg) is required by the ast-purity lint but is not on PATH   exit=1
```

The OLD-filter contrast proves the bug existed (not just that the lint works); the stripped-PATH contrast proves the no-silent-pass claim.

## Gates

- `actionlint .github/workflows/ci.yml` — clean (exit 0)
- `cargo fmt --check` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — no warnings
- `cargo nextest run --workspace --all-targets` — 1275 passed, 0 skipped
- `python3 scripts/bdd-tracked-lint.py` — ok
- `python3 scripts/mutants-skip-lint.py` — ok
- MSRV check — N/A (no Rust changed; workflow/shell only)

No hermetic CI test added: there is no natural home for a shell-lint unit test here, and the documented drill above satisfies AC2/AC3 — wiring one would be scope creep.

Closes #287

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1234,15 +1234,23 @@ jobs:
       - name: Install ripgrep
         # `ubuntu-latest` runner images do not ship ripgrep on every
         # generation (empirically observed via #261's `command -v rg`
-        # guard tripping on this PR's run). Three steps in this job
+        # guard tripping on PR #286's run). Three steps in this job
         # shell `rg` — the L2 grouped-import check, the L3.5 adapter-
         # name literal check (#161), and the L4 adapter-vocabulary
-        # check (#261). The L2 and L3.5 steps use the
-        # `if rg ...; then`/`|| true` patterns that silently treat a
-        # missing `rg` as "no matches found" and exit 0; the L4 step
-        # added in #261 surfaces this loudly via `command -v rg`.
-        # Explicit install closes the gap so all three layers
-        # actually execute. ~10-15s wall-clock overhead per CI run.
+        # check (#261). Explicit install guarantees presence so all
+        # three layers actually execute. ~10-15s wall-clock overhead
+        # per CI run.
+        #
+        # Defence in depth (#287): the explicit install is the
+        # empirical fix, but all three rg-shelling steps below ALSO
+        # carry an up-front `command -v rg` guard so that if this
+        # install ever regresses (or a future runner image drops
+        # ripgrep again), the gates fail loudly with `::error::`
+        # rather than silently treating a missing `rg` as "no matches
+        # found" and exiting 0. Prior to #287 only L4 carried the
+        # guard; L2 and L3.5 used the `if rg ...; then` / `|| true`
+        # silent-pass shapes, so they had been no-ops in CI on any
+        # ripgrep-less image since they shipped.
         run: sudo apt-get update && sudo apt-get install -y ripgrep
       - name: Reject AST-library imports in crap-core (leading-token form)
         # Scope: crap-core ONLY. oxc/swc imports are legitimate in
@@ -1271,14 +1279,29 @@ jobs:
         # `use {\n    syn::Item,\n};` which the leading-token regex
         # above misses (since the first token after `use` is `{`, not
         # the banned crate). Uses ripgrep's multiline mode (-U) +
-        # PCRE2 for `\b` word boundaries. ripgrep ships on
-        # ubuntu-latest's default toolchain.
+        # PCRE2 for `\b` word boundaries.
         #
         # The pattern bounds the match to a single `use { … }` block
         # via `[^}]*`, so nested braces inside a different scope
         # cannot drag a false-positive across statements.
+        #
+        # Tool-presence guard (same rationale as the layer-4 step
+        # below, #287): this step shells `rg`, and the
+        # `if rg ...; then` form treats a missing `rg` identically to
+        # "no matches found" — the lint would silently pass and exit
+        # 0. `ubuntu-latest` runner images do NOT ship ripgrep on
+        # every generation (empirically observed via #261's
+        # `command -v rg` guard tripping on PR #286's run), so the
+        # `Install ripgrep` step above guarantees presence; this
+        # up-front guard fails loudly if that install ever regresses
+        # or rg drops off the image, rather than letting the gate go
+        # silently dead.
         run: |
           set -euo pipefail
+          if ! command -v rg >/dev/null 2>&1; then
+            echo "::error::ripgrep (rg) is required by the ast-purity grouped-import lint but is not on PATH"
+            exit 1
+          fi
           if rg -nU --pcre2 'use\s*\{[^}]*\b(syn|quote|proc_macro2?|tree_sitter|swc|oxc)\b' crates/crap-core/src/; then
             echo "::error::crap-core must not import AST libraries (grouped-import form, e.g. \`use { syn::Foo };\`) — see CLAUDE.md and shaping.md"
             exit 1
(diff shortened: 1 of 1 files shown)


cmbays:
Additional validation — ran the hardened L3.5 `run:` block against **real `crates/crap-core/src`** (the scratch drills in the PR body proved the filter logic in isolation; this exercises the actual step on real source, which only CI otherwise runs):

**No false positive — CI stays green on merge.** The rg pre-pass hits 6 files; after the `#[cfg(test)]`-trim, 3 hits survive, all `///` doc-comment references:

```
crates/crap-core/src/domain/multi_lang.rs:59:    /// Stable wire identifier (e.g. `"crap4rs"`, `"crap4ts"`).
crates/crap-core/src/cli/mod.rs:677:           /// Adapter binary name (e.g., `"crap4rs"`, `"crap4ts"`). Drives
crates/crap-core/src/adapters/reporters/mod.rs:45: /// `env!("CARGO_PKG_NAME")` (`"crap4rs"`, `"crap4ts"`, …) from the
```

The tightened `^(\/\/|\/\*|\* |\*$)` filter correctly skips all three (they start with `//`) → step exits 0. The comment allowance is preserved on real source, not just synthetic fixtures.

**AC2 to-the-letter:** injected `*p = String::from("crap4rs");` into `crap-core/src/domain/multi_lang.rs` immediately **before** its `#[cfg(test)]` block, ran the hardened step:

```
L3.5 FIRES (exit 1) on:
crates/crap-core/src/domain/multi_lang.rs:371:    *p = String::from("crap4rs");
```

Then `git restore`d the file (working tree clean before and after — the commit contains only `ci.yml`). Bonus finding: a first attempt that *appended* the line at file-end did NOT fire — because it landed after the `#[cfg(test)]` marker, which the awk trim correctly excludes, confirming the test-exemption mechanic works as documented.

PR #360, opened 2026-05-30
feat(core+ci): #346 — ship crap.toml + wire scorecard-production to consume it

### LINKED ISSUE #346: config: ship crap-rs.toml + wire production scorecard to consume it
## Summary

Commit the root `crap-rs.toml` holding the shared analysis knobs, and wire the production scorecard CI job to consume it via auto-discovery — making `discover_config`/`load_config`/config-over-CLI-merge a real product surface, not just unit-tested. This is the core dogfood.

## Acceptance Criteria

- [ ] `crap-rs.toml` at repo root holds the shared analysis knobs: `threshold`/`preset`, `metric`, `exclude` globs, a `[views.ci]`-style preset, and (now that multi-root exists) the `src = [...]` array of the 3 prod crate roots
- [ ] The `scorecard-production` job (`.github/workflows/ci.yml`) consumes the config via auto-discovery — analysis knobs removed from the workflow YAML
- [ ] CI-presentation inputs (`gate-mode`/`comment-header`/`comment-preamble`/`html-report`) stay as action inputs (they are not analyzer config)
- [ ] The "action always passes `--src`" wrinkle is resolved (see Discovery)
- [ ] `crap-rs.toml` doubles as a worked reference, linked from README/docs

## Context

- Epic: #334
- Blocked-by: #345 (needs the unified name + discovery settled)
- Carries the #336 ADR constraint (`adr-multi-root-identity-base`): a config `src = [...]` with exactly ONE entry MUST stay src-relative (`len() == 1 ⇒ src-relative`) for envelope wire-compat — do NOT force repo-relative for all config-array runs.
- Code: `.github/workflows/ci.yml` (`scorecard-production` job, lines ~658-732), `crates/crap-core/src/adapters/config.rs`, the scorecard composite action (`.github/actions/scorecard/action.yml`)

## Discovery

- **The composite action always passes `--src` (default `.`)** → would override config `src`. Resolve: give the action a "defer to config" mode (omit `--src` when the input is unset) OR have the production job call the bare `crap4rs` binary directly. Analysis knobs (threshold/preset/etc.) dogfood cleanly via auto-discovery regardless; only `src`-via-config hits this.
- Which knobs move into config vs stay on the CLI/action (the analysis-vs-presentation line).
- Whether the production job keeps any explicit flags (e.g. `--coverage lcov.info`, which is per-run, not config).

cmbays:
## Summary

Commits a root `crap.toml` holding the production analysis knobs and wires the `scorecard-production` CI job to consume it via auto-discovery — making config-over-CLI-merge a real product surface (the core config dogfood), not just a unit-tested path.

## `crap.toml` (behavior-identical, values MOVED not changed)

Reproduces the `scorecard-production` job's prior inline knobs exactly:

| Knob | Value | Prior source |
|------|-------|--------------|
| `preset` | `"default"` (→ cognitive 15.0) | job's `threshold-preset: default` |
| `metric` | `"cognitive"` | crap4rs default (job set none) |
| `src` | the 3 prod crate roots | job's inline `src:` list |
| `exclude` | commented (inert) | job set none → effective empty |
| `[views.ci]` | worked reference | job passes no `--view` (does not gate) |

Linked from `README.md` as a real-world worked example alongside the exhaustive `crap.example.toml`.

## The "action always forwards a flag" wrinkle — **two** outliers, not one

`--src` was forwarded unconditionally (input default `'.'`). The locked fix: change `src`'s default to `''` and forward `--src` only when set — matching every other optional input.

While verifying, I found a **second outlier** the src-only framing missed: **`--threshold` was also forwarded unconditionally** (the resolve step substitutes the historical `15` when both `threshold`/`threshold-preset` are empty, then the analysis step always appends `--threshold`). Since crap4rs precedence is CLI `--threshold` > config `preset`/`threshold`, leaving it unconditional makes config-threshold **inert** — `src` would flow from config but threshold would be silently overridden by the wrapper's 15 (a dogfood that's a sham for threshold). I verified the gate reads `envelope.result.passed` (line 1066 — crap4rs computes it from the threshold it received), so suppressing the forward genuinely moves the value to config. Applied the **same conditional-forward treatment** to `--threshold` via a new `threshold_explicit` resolve-step output gating the forward in both the Run-analysis and Emit-annotations loops.

**Back-compat preserved** (verified by bash simulation): an explicit caller still gets `--threshold <n> --src <root>` exactly as before; only OMITTED inputs change behavior.

## `scorecard-production` job

Drops `src` and `threshold-preset` so `crap.toml` owns them via auto-discovery. CI-presentation (`gate-mode`/`comment-*`) and per-run (`coverage`) inputs STAY — not analyzer config.

## Behavior change (external action callers)

The `src` default change is a minor behavior change for any external caller that passed no `src` and relied on the wrapper scanning repo-root (`.`): they now get crap4rs's `["src"]` default. Documented as a framed callout in the action `README.md` (its natural changelog surface — the action is a CI surface, not a crate, so it has no per-crate CHANGELOG entry).

## Integration / blast radius (NOT "isolated" — broadly-read but value-matched)

Root `crap.toml` is auto-discovered by **every** scorecard-action caller that runs from the checkout root, not just the production job. Behavior stays identical because the config values **equal the prior action/binary defaults** (preset default → 15, cognitive, empty exclude): the other jobs pass an explicit `--src` (overrides config src) and no threshold (reads config 15 == old default). Verified empirically:

- Full `cargo nextest run --workspace --all-targets` — **1301 pass** with `crap.toml` present; wire-envelope canaries + self-scorecard unaffected (discovery is CWD-only → canaries run from `crates/crap-core`; they also pass explicit `--src`; the self-test calls library `analyze()` directly, never touching discovery).
- Binary dogfood from repo root: no-flags → config's 3 roots + threshold 15; explicit-`--src` → single-root + config threshold 15.
- `actionlint` clean; `zizmor --offline` no findings; back-compat bash sim confirms explicit callers unchanged.
- The one CI job using `annotations: true` sets `threshold: '8'` explicitly, so the annotations-step threshold forward is byte-identical for it.

## Gates

`cargo fmt --check` ✅ · `cargo +1.96.0 clippy --workspace --all-targets -- -D warnings` ✅ · `cargo nextest run --workspace --all-targets` (1301) ✅ · `cargo +1.93 check` (MSRV) ✅ · `bdd-tracked-lint` / `mutants-skip-lint` ✅ · `actionlint` ✅ · `zizmor --offline` ✅

AC2 ("gate must not fail / behavior-identical") is behavior-identical **by construction** (identical effective inputs); the real-coverage confirmation is this PR's own `scorecard-production` job running against the live LCOV artifact — watch it (and every scorecard job, per blast radius) go green.

## To ratify at merge

1. **`threshold` second-outlier / Path A** — the locked plan named `src` as "the lone outlier"; `--threshold` is a second one with the same root cause. I extended the same conditional-forward fix (not the bare-binary alternative). If you'd prefer to keep `threshold-preset: default` on the job and omit threshold/preset from config (behavior-safe, AC-partial), that's a one-line revert of the threshold half.
2. **`crap.toml` is now a global CI input** — changing its `preset`/`metric`/`exclude` later silently shifts every scorecard job that doesn't pass an explicit override, not just production.
3. **No CHANGELOG entry** — the change is in the composite action (a CI surface, not a crate); documented in the action README instead. Confirm that placement.

Closes #346

## Summary by CodeRabbit

* **Documentation**
  * Added documentation on configuration precedence and conditional input behavior when values are omitted.
  * Clarified that the repository demonstrates its own configuration via a production config file.

* **Chores**
  * Repository configuration now externalized to a dedicated config file with preset support.
  * Workflow updated to rely on config file values instead of hardcoded inputs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/breezy-bays-labs/crap-rs/pull/360?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


cmbays:
## Update: CI caught a real blast-radius interaction (now fixed) — `fe32b4d`

The first CI run failed the **`scorecard-smoke-ts`** job: root `crap.toml` is discovered by *every* adapter invocation from the checkout root — including **crap4ts** in the TS smoke + multi-lang smoke jobs — and the shared top-level `metric = "cognitive"` made crap4ts error (`metric 'cognitive' is not yet supported`; crap4ts only supports cyclomatic).

This is the dogfood working: local checks structurally could not exercise crap4ts-from-root-with-crap.toml, and CI surfaced it. Root cause: `metric` is inherently per-language (cognitive is crap4rs-specific), so it never belonged at the shared top level of a file both adapters read.

**Fix:** moved `metric = "cognitive"` into `[language.rust]` (which only crap4rs reads). crap4ts now finds no metric override → falls back to its cyclomatic default. Doubles as a live demonstration of D20's per-language override mechanism in the worked reference. Verified locally:
- crap4ts from root, explicit `--src`, no `--metric` → clean (no error)
- crap4rs from root → JSON envelope still `"metric": "cognitive"` (read from `[language.rust]`); production gate unchanged

`preset` stays top-level (a tier is metric-agnostic — valid for both adapters). `src` stays top-level (no per-language `src` field exists) with an added caveat comment: crap4ts discovers these Rust roots too, harmless only because every crap4ts CI job passes an explicit `--src` override. Full `cargo nextest run --workspace --all-targets` (1301) stays green.

This strengthens the "behavior-identical" claim: it's now confirmed across crap4rs AND crap4ts paths, with the cross-adapter footgun (one config, two readers) caught and scoped rather than assumed away.

cmbays:
## Bot review disposition (commit `a3abb61`)

Evaluated each finding against the code (primary source). Summary: 4 CodeRabbit findings applied (doc accuracy), 1 Gemini HIGH finding rejected as a false positive.

| Bot | Finding | Disposition |
|-----|---------|-------------|
| Gemini | **HIGH** — `metric` belongs under an `[analysis]` table (per "README line 909") | **Rejected — false positive.** There is no `[analysis]` table in the config schema (`crates/crap-core/src/adapters/config.rs` — `metric` is top-level / per-language), README has no line 909, and `[language.rust].metric` is empirically confirmed (crap4rs JSON envelope reports `"metric": "cognitive"`). Applying the suggested `[analysis]` table would break the config with an unknown key. |
| CodeRabbit | **Major** — `threshold-resolved` output still reports synthetic `15` when config owns the threshold | **Applied (doc).** Documented that the output reflects the action's resolved *input*, not the analyzer's effective gate threshold; config-driven consumers should read the envelope. No current consumer is misled (the production job ignores this output; the preset spot-checks all pass explicit thresholds), and the "always a number" contract is intentionally preserved. |
| CodeRabbit | Minor — "cognitive 15" fallback wording implies the TS path uses Rust calibration | **Applied.** Reworded to "each adapter keeps its own metric calibration" (crap4rs cognitive, crap4ts cyclomatic). |
| CodeRabbit | Minor — Inputs table still lists `src` default as `.` | **Applied.** Updated to `''` + a note that omitted `src` is no longer forwarded. |
| CodeRabbit | Minor — README claims the production job "consumes" `[views.ci]` | **Applied.** Reworded: the preset is a worked reference that only shapes a report under `--view ci` (the job passes none). |

CodeRabbit's `[views.ci]` finding was already marked "✅ Addressed in commit fe32b4d" by CodeRabbit itself.

PR #374, opened 2026-06-03
feat(core+cli): #278 — configurable missing-coverage policy (pessimistic|optimistic|skip)

### LINKED ISSUE #278: matching: configurable missing-coverage policy (pessimistic | optimistic | skip)
## Summary
Replace the hardcoded pessimistic behavior in `match_functions` (functions in files absent from LCOV scored as 0% covered → CRAP = c² + c) with a configurable policy. Unlocks deleting the feature-gated-shim workaround currently in use to dodge self-CRAP inflation.

## Acceptance Criteria
- [ ] New `--missing-coverage-policy {pessimistic|optimistic|skip}` flag + `crap4rs.toml` key
- [ ] Default = `pessimistic` (current behavior, no breaking change)
- [ ] `optimistic`: functions in files missing from LCOV scored at 100% coverage (CRAP = complexity) — appropriate when running a scoped test slice during local dev
- [ ] `skip`: functions in files missing from LCOV excluded entirely from the result (don't appear in summary, scorecard, or function list)
- [ ] `functions_no_coverage` diagnostic counter remains accurate under all three policies (count of functions whose file was missing from LCOV, regardless of how they were handled)
- [ ] Wire-envelope records the chosen policy in metadata so baselines + deltas stay coherent across policy changes (avoid silent CRAP inflation/deflation when policy changes between baseline and current runs)
- [ ] Snapshot tests for all three policy outputs

## Context
- `crap-core/src/domain/matching.rs:36-46` hardcodes the pessimistic branch via `zero_coverage()` at line 129 — `line_data.get(&comp.identity.file_path)` returns `None` → forces 0% coverage
- This already cost us a workaround: feature-gated modules score `c² + c` in the self-CRAP run because the default-features `lcov.info` has no record of them. The workaround was to keep those shims artificially thin to dodge inflation. A policy knob lets us delete the workaround and run self-CRAP with `optimistic` policy instead
- Surfaced via external tool comparison — another tool in this space exposes the same `pessimistic|optimistic|skip` policy enum for this case
- LCOV-missing causes: feature flags (`#[cfg(feature = "x")]` files not compiled in the coverage build), untested files, scoped per-package coverage runs (e.g. `cargo llvm-cov -p crap-core` won't have `SF:` records for `crap4rs` source files even though they exist on disk)

## Discovery
- Should `optimistic` policy emit a stderr diagnostic (or surface a counter) so users see how many functions were treated as 100% covered without being tested?
- Self-CRAP workflow follow-up: should we switch the repo's self-CRAP run to `optimistic` policy and delete the thin-shim workaround in a subsequent PR? File as follow-up child issue once this lands
- Interaction with `--src` filter: if a file is excluded by `--src`, the function never makes it into `complexities` and so never hits this branch. Confirm the policy only fires for files passing `--src` that are missing from LCOV (expected behavior)
- Baseline/delta coherence: if baseline was generated with `pessimistic` and current with `optimistic`, every previously-zero-coverage function will appear as a massive improvement. The wire-envelope metadata should at minimum let the delta engine WARN on policy mismatch; AC currently just records it — confirm whether warn/error is right

cmbays:
## Summary

Replaces the hardcoded pessimistic scoring for functions whose source file is **entirely absent from the coverage data** with a configurable `--missing-coverage-policy {pessimistic|optimistic|skip}` flag (and matching `missing_coverage_policy` config key). The default is `pessimistic` — byte-identical to today's behavior.

| Policy | Behavior |
|---|---|
| `pessimistic` (default) | score at 0% coverage (CRAP = c² + c) — never hides risk |
| `optimistic` | score at 100% coverage (CRAP = c) — for a scoped local test slice |
| `skip` | omit those functions from the report entirely |

The knob lives in `crap-core` (domain enum + CLI + matching), so **both adapters inherit it** through the shared `analyze` pipeline.

## What changed (commit-by-commit)

1. **domain + core mechanism** — `MissingCoveragePolicy` enum; `match_functions` takes the policy; `functions_no_coverage` is now derived from `all_complexities` (not the matched vec) so it stays accurate under `skip`, and the partition `debug_assert` re-anchors to `all_complexities` as a real check.
2. **CLI flag + config key** — `--missing-coverage-policy` (`Option<…>` so an unset flag never clobbers a config-set policy past `..Default`), `missing_coverage_policy` config key, cascade CLI > config > default. Regenerated `crap.example.toml` + `crap.schema.json`.
3. **envelope + baseline mismatch warn** — the JSON envelope records the policy, **elided when `pessimistic`** so every existing envelope (incl. the self-referential `wire_envelope_crap4rs` snapshot) stays byte-identical; the baseline loader reads it with `#[serde(default)]` (absent ⇒ pessimistic). A delta run **warns when its policy differs from the baseline's** — the `pessimistic` → `skip` case is the sharpest (those functions are present in the baseline, absent now, so the delta reads them as deletions).
4. **end-to-end 3-policy snapshot + docs** — a single function whose file is absent from coverage scores three ways through the real pipeline; crap4rs README documents the flag.
5. **refactor below strict-8** — the policy dispatch pushed `match_functions` to cognitive 8 (CRAP 8.0 at 100% cov — passing, but at the boundary). Extracted `coverage_for_function`; `match_functions` is now cognitive 1, the helper 4.

## AC coverage (#278)

- [x] `--missing-coverage-policy` flag + config key; default `pessimistic`
- [x] `optimistic` → 100% (CRAP = c); `skip` → omitted from summary/scorecard/list
- [x] `functions_no_coverage` accurate under all three policies
- [x] policy recorded in the wire envelope (baseline/delta coherence) + mismatch warning
- [x] tests for all three policy outputs (matching unit, envelope serialize, baseline round-trip, and an end-to-end snapshot)

## Verification

- `cargo fmt --check`, `cargo +1.96.0 clippy --workspace --all-targets -- -D warnings`, `cargo +1.93 check --workspace --all-targets` (MSRV) — all clean
- Full workspace: **1364 tests pass**; `scripts/mutants-skip-lint.py` + `scripts/bdd-tracked-lint.py` green
- **strict-8 self-gate** (the `scorecard-production` job): ran locally over the three prod crates with real `cargo llvm-cov` coverage — exit 0, zero functions over CRAP 8. The two touched functions land at CRAP 1.0 (`match_functions`) and 4.0 (`score_and_assemble`), with the new `coverage_for_function` at 4.0.

## Note on crap4ts

crap4ts inherits the knob for free via shared `crap-core`. Its Istanbul/V8 coverage is file-path-keyed the same way LCOV is, so a TS function whose file is absent from the coverage map is scored per the policy identically — but I did **not** separately exercise the TS path against a coverage-absent file in this PR; the parity is by construction (one shared `match_functions`), not separately validated end-to-end here.

Closes #278

## Release Notes

* **New Features**
  * Added `--missing-coverage-policy` CLI option to control how functions are scored when source files are absent from coverage data, with three modes: pessimistic (default, 0%), optimistic (100%), or skip (omit from report).
  * Added corresponding `missing_coverage_policy` configuration option in TOML files.
  * Policy selection is recorded in JSON envelopes for baseline comparisons; mismatched policies between runs trigger a warning.

* **Documentation**
  * Updated documentation with guidance on the new missing coverage policy option and its behaviors.

### DIFF
--- a/crap.example.toml
+++ b/crap.example.toml
@@ -15,6 +15,11 @@ threshold = 15.0
 # or `"cyclomatic"`.
 metric = "cognitive"
 
+# How to score a function whose source file is entirely absent from
+# the coverage data: `"pessimistic"` (default — 0% covered),
+# `"optimistic"` (100% covered), or `"skip"` (omit from the report).
+missing_coverage_policy = "optimistic"
+
 # Source root(s) the analyzer walks. Accepts a single string
 # (`src = "crates"`) or an array (`src = ["crate-a", "crate-b"]`).
 # A single root stays src-relative; multiple roots are keyed
--- a/crap.schema.json
+++ b/crap.schema.json
@@ -28,6 +28,13 @@
         "null"
       ]
     },
+    "missing_coverage_policy": {
+      "description": "How to score a function whose source file is entirely absent from\nthe coverage data: `\"pessimistic\"` (default — 0% covered),\n`\"optimistic\"` (100% covered), or `\"skip\"` (omit from the report).",
+      "type": [
+        "string",
+        "null"
+      ]
+    },
     "output": {
       "description": "Output-shaping settings (`[output]`): annotation cap, scorecard\ntitle, and subtitle.",
       "$ref": "#/$defs/OutputSchema"
--- a/crates/crap-core/src/adapters/baseline.rs
+++ b/crates/crap-core/src/adapters/baseline.rs
@@ -16,7 +16,7 @@
 //! column-keyed). Future schema bumps will need an explicit
 //! migration path.
 
-use crate::domain::types::{AnalysisDiagnostics, AnalysisResult};
+use crate::domain::types::{AnalysisDiagnostics, AnalysisResult, MissingCoveragePolicy};
 use crate::ports::ParseDiagnostic;
 use serde::Deserialize;
 use std::fs::File;
@@ -55,6 +55,13 @@ struct BaselineEnvelope<P: ParseDiagnostic> {
     result: AnalysisResult,
     #[serde(default)]
     diagnostics: Option<AnalysisDiagnostics<P>>,
+    /// Policy the baseline was generated under. Absent on envelopes
+    /// emitted before the field existed, and on any pessimistic run
+    /// (the emitter elides the default) — both deserialize to
+    /// `Pessimistic` via `Default`, the load-bearing "absent ⇒
+    /// pessimistic" convention the delta mismatch warning relies on.
+    #[serde(default)]
+    missing_coverage_policy: MissingCoveragePolicy,
 }
 
 /// What the loader returns to callers. The fields are all the metadata
@@ -69,6 +76,10 @@ pub struct BaselineSnapshot<P: ParseDiagnostic> {
     pub tool_version: String,
     pub timestamp: String,
     pub diagnostics: Option<AnalysisDiagnostics<P>>,
+    /// Policy the baseline was generated under (`Pessimistic` when the
+    /// envelope omits the key). The delta path compares this to the
+    /// current run's policy and warns on a mismatch.
+    pub missing_coverage_policy: MissingCoveragePolicy,
 }
 
 /// Errors raised while loading a baseline envelope.
@@ -147,6 +158,7 @@ pub fn load<P: ParseDiagnostic>(path: &Path) -> Result<BaselineSnapshot<P>, Base
         tool_version: envelope.tool_version,
         timestamp: envelope.timestamp,
         diagnostics: envelope.diagnostics,
+        missing_coverage_policy: envelope.missing_coverage_policy,
     })
 }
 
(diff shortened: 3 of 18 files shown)


cmbays:
## Bot review + CI status

**Bot review — clean, no findings to address:**
- **Gemini**: "no feedback to provide."
- **CodeRabbit** (CHILL profile): 0 actionable comments, 0 inline comments. All 5 pre-merge checks passed — Title ✓, Linked Issues (all #278 ACs met) ✓, Out of Scope ✓, Docstring Coverage 100% ✓.

**One CI fix pushed** (`0ff7dbf`): the first run's `cargo doc (warnings as errors)` job failed because `match_functions` (public) linked to `coverage_for_function` (private) — `rustdoc::private-intra-doc-links` is an error under `-D warnings`, though it only warns on a bare local `cargo doc`. Fixed by stating the behavior directly instead of linking the private helper. No behavior change.

Everything else on the first run was green, including **`Scorecard — production crates`** (the strict-8 self-gate) — verified locally too with real `cargo llvm-cov` coverage: exit 0, zero functions over CRAP 8, the touched functions at CRAP 1.0 / 4.0.

Re-running CI on the doc fix now.

PR #441, opened 2026-06-20
ci: fix fuzz-nightly invocation (cargo bolero test + profile.fuzz + real target names)

cmbays:
## Summary
Follow-up fix: the first-fire of `fuzz-nightly.yml` (dispatched after #440 merged) **failed** — the shaping probe had assumed the wrong cargo-bolero CLI. Three corrections, all validated empirically against cargo-bolero 0.13.4 (`cargo bolero list` + a local nightly run):

1. **`cargo bolero test`**, not `cargo bolero fuzz` — 0.13 has no `fuzz` subcommand (`error: Found argument 'fuzz' which wasn't expected`). `test` drives the engine.
2. **`[profile.fuzz]` added to the workspace `Cargo.toml`** — cargo-bolero builds `--profile fuzz` and errors `profile 'fuzz' is not defined` without it. bolero's recommended shape (`inherits = "dev"`, opt). The stable per-PR nextest lane never uses this profile.
3. **Real target names** — the matrix used directory names, but bolero targets are the `check!` fn names. `cargo bolero list` confirms 5: `fuzz_syn_walker_cyclomatic` + `fuzz_syn_walker_cognitive` (crap4rs); `fuzz_oxc_walker`, `fuzz_istanbul_raw`, `fuzz_istanbul_structured` (crap4ts). Added a `dir` matrix field for the artifact path.

## Validation
- `cargo bolero list` → confirmed the 5 target names + that `[profile.fuzz]` resolves the build.
- Local nightly run reached **linking** with the corrected invocation (flags all parse + build), then hit a **macOS-only** ASan ld error (`initializer pointer has no target`) — CI is ubuntu (ASan links there).
- **The corrected workflow was dispatched on this branch** ([run 27857176312](https://github.com/breezy-bays-labs/crap-rs/actions/runs/27857176312)) to validate the real fuzz run on Linux — merging once it's green.
- actionlint + zizmor clean; `cargo metadata` confirms the profile parses (normal builds unaffected).

This is the first-fire validation flagged on #440. No new behavior on the per-PR path.

## Summary by CodeRabbit

* **Chores**
  * Expanded the nightly fuzz testing workflow with additional targets and per-target corpus/crash directories, and improved crash artifact uploads to include the full per-target output.
  * Added a dedicated `fuzz` Cargo profile with optimized, non-incremental build settings.
* **Tests**
  * Updated the Cucumber-based test runners to apply default CLI configuration before executing the tagged scenarios.

### DIFF
--- a/.github/workflows/fuzz-nightly.yml
+++ b/.github/workflows/fuzz-nightly.yml
@@ -76,11 +76,15 @@ jobs:
           fi
       - name: Fuzz ${{ matrix.target.test }}
         # `cargo bolero test <target>` is the fuzz command (cargo-bolero 0.13
-        # has no `fuzz` subcommand; `test` drives the engine). Target names +
-        # the `--sanitizer/--engine/-T` flags + the `[profile.fuzz]` requirement
-        # were validated empirically via `cargo bolero list` + a local nightly
-        # run. The job's default toolchain is nightly (dtolnay above), so no
-        # `--toolchain` flag is needed; `--sanitizer address` ⇒ libFuzzer+ASan.
+        # has no `fuzz` subcommand; `test` drives the engine). `--toolchain
+        # nightly` is REQUIRED: `--sanitizer`/libFuzzer need `-Zsanitizer`
+        # (nightly-only), and the repo's `rust-toolchain.toml` pins
+        # `channel=stable` — which overrides dtolnay's installed-nightly
+        # default, so without this flag cargo-bolero runs stable rustc and
+        # `-Z` is rejected. (The dtolnay step above installs the nightly this
+        # flag selects.) Subcommand + target names + `[profile.fuzz]` +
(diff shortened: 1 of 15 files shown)


cmbays:
Superseded. First-firing the lane proved that `cargo bolero test` has no test-target scoping — it drags the whole crap4rs cucumber suite (which fails under ASan) into every fuzz run — so `.with_default_cli()` only changed the failure mode, not the root cause. Splitting into focused follow-ups instead:

- **#442** — fixes the real upstream-oxc walker panic this lane found on first fire (+ committed regression seed).
- **#444** — defers the lane: disables the weekly schedule, drops the crap4rs targets, keeps the 3 working crap4ts targets on `workflow_dispatch`.
- **#443** — tracks restoring the full scheduled lane via a dedicated fuzz crate (the idiomatic cargo-fuzz pattern that gives real test-target scoping).

Closing without merge; the branch stays for reference.

PR #454, opened 2026-06-24
ci: encoded-defect lint — anchor deliberately-wrong values to a tracking issue

cmbays:
Release-hardening series, part 3.

## What

A CI lint + lefthook hook that **fails on a deliberately-wrong / deferred value** frozen into a test, snapshot, or fixture **unless it carries a tracking anchor** — `tracked: crap-rs#<n>` (active) or `resolved: crap-rs#<n>` (historical/post-mortem).

## Why

The rc.4 release shipped a wire envelope with `language` frozen to the wrong value (`"rust"` for the TS adapter) under a "wrong-by-design pending a later fix" comment — and because a shape-lock **snapshot was the oracle**, the deferred defect sailed green through every PR until publish. This makes that deferral class accountable (the exclusions discipline, applied to values). Corollary documented in AGENTS.md: a snapshot is a *tautological* oracle; use an explicit `assert_eq!` for knowable-correct values.

## False-positive safety (this is a build-failing heuristic lint)

- Matches only **unambiguous** markers — `wrong-by-design`, `known-wrong`, `wrong-on-purpose`, `deliberately wrong`, `intentionally wrong`. The discriminator is *wrong*; bare **"by design"** (8 benign uses in the tree) is **not** matched.
- Dual `tracked:`/`resolved:` anchor within a 6-line window.
- Scoped to source/test/snapshot/fixture files under `crates/` (`.rs .snap .json .ts .toml .feature`); **excludes `scripts/` and `*.md`** — the lint and these docs name the markers in their own definitions (self-match avoidance).
- **Adversarial whole-tree FP-hunt** at authoring flagged *only* the known post-mortem doc block in `wire_envelope_crap4ts.rs` — now anchored `resolved: crap-rs#188` (metric flip) / `crap-rs#450` (language flip), both shipped. Zero other matches.
- `--self-test` (13 cases incl. bare-"by design", benign "wrong"/"design", anchor-outside-window) guards the rule logic; runs in CI before the live check.

**Honest limitation** (in the AGENTS.md section): it's a heuristic text matcher — it enforces that a *labeled* deferral is *anchored*; it can't detect a deferred value that carries no marker at all.

## Validation

fmt + actionlint + zizmor clean; lefthook validates; `wire_envelope_crap4ts` test passes (doc-only edit, no snapshot change); lint self-test 13/13; lint clean on the tree.

Part of the release-hardening series (crap-rs#448-adjacent). Decision captured in the release-process ADR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1888,6 +1888,43 @@ jobs:
           BASE_SHA: ${{ github.event.pull_request.base.sha }}
         run: python3 scripts/version-edit-lint.py
 
+  encoded-defect-lint:
+    # Mechanical enforcement of AGENTS.md "Encoded-defect discipline": a
+    # deliberately-wrong / deferred VALUE frozen into a test, snapshot, or
+    # fixture must carry a tracking anchor — `tracked: crap-rs#<n>` (active)
+    # or `resolved: crap-rs#<n>` (historical/post-mortem). The rc.4 release
+    # shipped a wire envelope whose `language` was frozen to the wrong value
+    # under a "wrong-by-design pending a later fix" comment, and because the
+    # snapshot WAS the oracle the deferred defect sailed green to publish.
+    # This gate makes that deferral class accountable — same discipline as
+    # `~/.claude/rules/exclusions.md` (documentation rots; CI doesn't).
+    #
+    # Matches only unambiguous "deliberately wrong" markers (wrong-by-design
+    # / known-wrong / wrong-on-purpose / deliberately|intentionally wrong),
+    # NOT bare "by design" (a legitimate design statement). Scoped to
+    # source/test/snapshot/fixture files under `crates/` (excludes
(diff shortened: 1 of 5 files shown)


cmbays:
Addressed all 3 Gemini findings in db0df78 (threads resolved):
- **marker regex (L73):** separator broadened `[-\s]` → `[-_\s]+` so `wrong_by_design` / `wrong  by  design` (underscore + multi-space) match too, without widening the vocabulary. +3 self-test cases (16/16).
- **dead scripts/ filter (L129):** removed — the scan is scoped to `crates/` via `git ls-files crates/`, so scripts/ and root `*.md` are excluded by scope already (clarified in a comment).
- **invalid-UTF-8 (L140):** `open(..., errors="replace")` so a file with stray invalid-UTF-8 bytes is scanned rather than silently skipped.

Whole-tree FP-hunt still flags nothing (the post-mortem doc is anchored).

PR #455, opened 2026-06-24
ci: shift-left npm packaging verification onto the release-plz PR

cmbays:
Release-hardening series, part 2 (the final PR).

## What

The npm pack + install-gating check — the smoke that catches a packaging defect like the crap4ts 2.0.0-rc.1 stray `libc` field (which made `npm install` reject the platform with EBADPLATFORM) — ran **only post-merge** in release-plz.yml's per-platform `build-crap4ts-cdylib` job. This extracts it into a shared script and adds a **pre-merge** job on the release-plz PR so the defect fails **left of the irreversible publish merge**.

- `scripts/verify-npm-packaging.sh` — single source of truth: pack the package, install the tarball into a throwaway consumer, assert the `analyze` export (the os/cpu/libc install-gating a path-`require()` of the `.node` can't exercise).
- `release-plz.yml` `build-crap4ts-cdylib` (post-merge, per-platform) now calls the script. The `npm install` of a local tarball moved into the script (out of zizmor's view), so the former inline `# zizmor: ignore[adhoc-packages]` is dropped.
- `ci.yml` `verify-npm-packaging` job (pre-merge): builds the native `linux-x64-gnu` cdylib + renames to the napi `.node` filename (mirrors the matrix row) + runs the shared script. Gated to `release-plz-*` PRs — the deliberate dev→prod gate.

## Scope note

`cargo publish --dry-run` is deliberately **not** added pre-merge: it can't resolve for crates depending on the not-yet-published crap-core (crates.io publish ordering; release-plz handles it at real publish), and `semver_check = true` already gates the release PR. So the genuinely-shiftable release-only gate here is the npm packaging check.

## Validation

shellcheck + actionlint + zizmor clean on both workflows; setup-rust inputs confirmed; build/rename values mirror the existing matrix.

Part of the release-hardening series. Decision captured in the release-process ADR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1925,6 +1925,52 @@ jobs:
       - name: Reject unanchored deliberately-wrong markers
         run: python3 scripts/encoded-defect-lint.py
 
+  verify-npm-packaging:
+    # Shift-left of the npm packaging + install-gating gate. The same
+    # check runs post-merge in release-plz.yml's `build-crap4ts-cdylib`
+    # (per-platform matrix); here it runs on the release-plz PR (native
+    # linux-x64-gnu) BEFORE the irreversible publish merge, so a packaging
+    # defect — e.g. a stray os/cpu/libc field that makes `npm install`
+    # reject the platform with EBADPLATFORM, as crap4ts 2.0.0-rc.1 shipped
+    # — fails on the release PR instead of reaching the registry. Single
+    # source of truth: `scripts/verify-npm-packaging.sh`.
+    #
+    # Gated to the release-plz PR (the deliberate dev->prod gate). The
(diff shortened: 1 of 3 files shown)


cmbays:
Addressed both Gemini findings in a908451 (threads resolved):
- **temp cleanup + ls-parse:** added `trap 'rm -rf "$work"' EXIT` and replaced the `ls` parse with a glob array (`tarballs=("$work"/crap4ts-*.tgz)`) guarded by a `-f` check.
- **cwd / trap:** the consumer `npm init`/`install`/`node -e` now run in a subshell, so the main shell cwd is unchanged and the EXIT trap can remove `$work`; a non-zero exit inside still propagates via `set -e`.

shellcheck clean; the pack+install+analyze mechanism re-verified locally on the native platform.

